Issue #2, opened 2016-10-14
Remove RWD


arottersman:
Remove all Rapid Watershed Delineation related code

PR #7, opened 2016-10-14
Remove RWD

arottersman:
Remove RWD from the project
We no longer setup for it in deployment, have tasks for it on the backend, or show the dropdown for it in the frontend.

_Delineate Watershed Gone!_
screen shot 2016-10-14 at 5 14 00 pm

### Testing
Assuming you've already experienced the joy of `vagrant up`, grab this branch and confirm that on `localhost:8000` the delineate dropdown is gone from the main toolbar.

In your fave IDE search the project for things like "rwd" and "delineat". There shouldn't be any references to that sort of thing. 

Connects #2


kellyi:
I set up the app VM and confirmed the option's now gone:

screen shot 2016-10-17 at 9 56 55 am

kellyi:
No references to rwd or delineation when searching through Atom.

Looks good to me. +1

arottersman:
Thanks, Kelly!

deployment/cfn/worker.py
```diff
@@ -342,7 +342,7 @@ def get_cloud_config(self):
         return ['#cloud-config\n',
                 '\n',
                 'mounts:\n',
-                '  - [xvdf, /opt/rwd-data, ext4, "defaults,nofail,discard", 0, 2]\n'  # NOQA
```
mmcfarland: I believe the configuration for this whole mount should be removed, not just the path.  Since it's the only mount we made otherwise, the whole mounts section should be excised as well.

src/mmw/js/src/analyze/tests.js
```diff
@@ -65,7 +65,11 @@ function testAnalysisType(type) {
                 actualTableRows = $('table tbody td').map(function() {
                     return $(this).text().trim();
                 }).toArray();
+            console.log("expected rows => " + expectedTableRows);
```
mmcfarland: These debugging statements should be removed, the tests seem to be asserting the equality.

arottersman: Yikes, this was me testing something for some breaking tests on mmw- They must have ended up  in my patch file.

deployment/cfn/worker.py
arottersman: Since this has already been merged, what's the best way to add the fixups? Should I open a new PR?

Issue #4, opened 2016-10-14
Remove Ability To Add Overlays


arottersman:
(from the scope of work doc) Remove any map overlays and reference layers, including stream network overlays, sensor observations integration, watershed and political boundaries, and existing land use and soil type raster overlays

PR #8, opened 2016-10-14
Remove tiler, layer control, layer settings, and related Ansible & deployment code

kellyi:
This PR removes code related to the tiler, layer control, and layer settings from the app and deployment code. In addition to dropping the tiler VM and the related Ansible and Vagrant settings, it also cuts the layer settings file and removes references to the layer settings and `DRB_PERIMETER` constant from Django, drops the custom layer control and the code to create the layers from the frontend, and cuts all the observations code including stuff on the frontend and in Nginx

With the change, we've dropped the basemap, too, and we'll add it back using a standard leaflet layers control:

screen shot 2016-10-17 at 11 12 09 am

**Testing**
- get this branch, then `vagrant up --provision app`, and once that's done run `./scripts/bundle.sh`
- visit `localhost:8000` and open the browser console
- verify that you see the above screen (which doesn't include the layers control), and that there aren't any errors in the console.

Note that the "Loading..." dropdown is for the boundary layer selector which will be removed in #3 

Connects #4


arottersman:
Do we still need this condition for Windshaft in the logstash setup? 
https://github.com/project-icp/bee-pollinator-app/blob/cb69b69c5ebba1a578846a6c11e7900116594864/deployment/ansible/roles/model-my-watershed.logstash/templates/logstash.conf.j2#L15

arottersman:
Finally able to test this...things took awhile because every now and then when I reload my app VM, it fails with `app: Warning: Authentication failure. Retrying...`, does this happen to anyone else?

Anyway, looks good!

kellyi:
Nice catch on the logstash windshaft code. Let me cut that and push another commit.

kellyi:
Cut those lines, adjusted the casing on the commit messages, and pushed an update. 

I haven't seen the authentication failure error with the VM, but I've seen it fail in a few other ways. Since there's no data loaded into this one yet, I think you could `vagrant destroy app` then `vagrant up` if worse came to worse.

arottersman:
Awesome +1

Yeah, usually destroying the VM fixes it for a little bit (although it's painful even without any data loaded yet)

kellyi:
Thanks for taking a look at this!

PR #11, opened 2016-10-17
Remove boundary layers, selector, and endpoints

kellyi:
This PR removes the boundary layers, selector, and Django endpoints from the app. Note that only commits 	67bbae0 and following are from this PR; the rest are from #8 

The changes here include:
- removing code related to the boundary layers and tests from Django
- cutting the boundary layer selector from the frontend, along with code to retrieve and display the boundary overlays on the map and code to display labels on mouseover
- cutting boundary layer-specific code from the templates
- updating the README to remove a reference to loading the boundary layers data
- adjusting the geocoder search icon to place it back in the vertical middle of the geocoder search bar
- updating a front-end test to change the expectation that there are 3 draw tools to an expectation that there's 1 (since we removed RWD and the Boundary Layer Selector, only the free draw/sq km dropdown is left

**Testing**
- get this branch, `vagrant up --provision app`, `./scripts/bundle.sh` and then visit `localhost:8000`
- open the browser console and verify that the app loads (minus a basemap) and no longer shows the boundary selector dropdown. also verify that you don't see any frontend errors in the console
- `./scripts/bundle.sh --tests` then `./scripts/testem.sh` and visit `localhost:7357` and verify that the tests all pass

Connects #3


lewfish:
We typically begin commit titles with capital letters.

src/mmw/apps/modeling/views.py
```diff
@@ -447,32 +447,6 @@ def _construct_tr55_job_chain(model_input, job_id):
...
-        row = cursor.fetchone()
-
-        if row:
-            geojson = json.loads(GEOSGeometry(row[0]).geojson)
-            return Response(geojson)
-        else:
-            return Response(status=status.HTTP_400_BAD_REQUEST)
-
-
-@decorators.api_view(['GET'])
```
lewfish: I think these two lines weren't supposed to be removed as part of this PR.

README.md
```diff
@@ -47,18 +47,12 @@ $ ./scripts/manage.sh migrate
...
+To load or reload data, from an `app` server, run (`scripts` is not mounted by default to the VM, you may need to copy the file over):
 
 ```bash
 $ ./scripts/setupdb.sh -b
 ```
 
-The same script can be used to load the stream network data:
-
-```bash
-$ ./scripts/setupdb.sh -s
```
lewfish: I think you meant to remove the `-b` option above since that controls boundary data. Also, you should remove these options from the script itself.

kellyi: Good catch. I cut this one too. I had removed most of the options from the `setupdb.sh` script in #8 but left one for loading `bee_data` as a model to use when we start adding new tables.

kellyi:
Pushed an update which updates the README and fixes the casing on the commit messages. The commits for this PR now begin with 7bddae3; everything else is from a prior PR.

kellyi:
Rebased this on `develop`.

lewfish:
+1, tested

kellyi:
Thanks for taking a look at this!

Issue #5, opened 2016-10-14
Remove gwlf-e and tr-55


arottersman:
Leave the model execution flow for one of them (but have it return some dummy json instead of calling the actual model); that way, when we have the bee model we can just replace stuff where applicable.

PR #12, opened 2016-10-17
Remove Most Of GWLF-E and TR55

arottersman:
Connects #5 

The Bee Pollinator won't need the TR-55 model or the GWLF-E model, but we want to keep enough of their execution flow to be able to easily find+replace/pop in the bee pollinator model.  

- Removes GWLF-E+Mapshed from the backend entirely
- Leaves TR55 throughout, but adds TODO comments in spots I thought would be be good starting points for switching out TR55 for the bee model
- Leaves GWLF-E+Mapshed in a few spots in the frontend in case we want to reuse any of the user input code/validation that's there. I've tried to comment `TODO + reason for leaving line` everywhere I left something pertaining to GWLF-E. On both the frontend and backend there's `gis_data`  which only GWLFE-E used - I wasn't sure how relevant it could be to the bee model, so I left it, too. 

### Testing
After pulling this branch you should still be able to build the project and see the mapless app at `localhost:8000`.

- If you draw a shape, there should be a bunch of network errors as the frontend tries to pull the analysis and tr55 endpoints.
- If you click the model dropdown, only `Site-Storm Model` should appear.


arottersman:
Pushed a fixup that removes some gwlf-e related code that I missed (the task that loads the DRB point source data, and the `gwlfe` directory in `js/src/modeling`)

jtarricone:
+1 Tested. There are some flake and JS linter warnings but probably nothing yet worth spending time on while we're still gutting the project.

arottersman:
Thanks for taking a look, Joe!

Issue #14, opened 2016-10-17
Remove Analyze View


arottersman:
Remove the analyze tabs but leave the slide out panel.

PR #15, opened 2016-10-18
Gut Analysis View

arottersman:
Connects #14 

Currently branched off of #12, the relevant commits are from `c197da5` on

- Removes all the frontend analyze tests
- Removes all the calls on the frontend for specific analyze tasks, eg `analyze/pointsource`, and each task's corresponding tab
- Leaves the `land` tab, but just shows the AOI info instead of the histogram
- Removes some Django templates `analyze.html` and `model.html` that I couldn't find being used anywhere

### Testing
Pull this branch. Assuming your VMs are running...
`./scripts/bundle.sh`

 On `localhost:8000`...
- Verify that after drawing a shape, you see only one tab in the analyze panel ("Land")
- There should still be the usual info about the selected area in the analyze panel
- Verify there are no network errors in the console


lewfish:
The code looks good. I was just wondering if any of the analysis related functions need to be removed in https://github.com/project-icp/bee-pollinator-app/blob/develop/src/mmw/apps/modeling/tasks.py Maybe we need to keep them though because they'll be used as part of the model run.

lewfish:
When you draw an AOI and then proceed to the analyze view, the outline of the area disappears, and also isn't there when going back by clicking the Draw button. Was this intentional?

arottersman:
Re: Disappearing AOI - not intentional. I've been combing the code that we've taken out and stepping through the code that we still have in and I can't find what's making it disappear... 

Would taking out all the basemaps affect making/applying the mask from the AOI? (It's actually been disappearing since a few merges ago)

kellyi:
Does the above happen on `develop` too? If so, I may have unwittingly dropped the code which handles highlighting the drawn aoi when I removed support for the boundary layers.

arottersman:
Yeah, it happens on develop. [This ](https://github.com/project-icp/bee-pollinator-app/blob/bfefe7b890a4576eae0142ae36bd6fa761eb6707/src/mmw/js/src/core/views.js#L448) seemed to me to be the relevant bit, but it's all still there.

arottersman:
We did remove `addAdditionalAoiShapes` from the same `core/views.js` file, but when I added it back in, the AOI still disappeared.

kellyi:
Here's what I cut from the JS code in #11: https://github.com/project-icp/bee-pollinator-app/pull/11/commits/0cab48df9bb41990e8c52625a5ca87eca54c2637

arottersman:
@kellyi Doesn't look like you took out anything that'd cause this in that commit

arottersman:
Aha! https://github.com/project-icp/bee-pollinator-app/commit/18bc8828126c5df38a468a034c0f7e1a41b21c2c#diff-288365c190603e65aeebe8aab8949361L239

arottersman:
Should I go ahead and merge this, or would you like to test that the AOI sticks around in the analyze view?

src/mmw/js/src/core/views.js
```diff
@@ -235,10 +235,9 @@ var MapView = Marionette.ItemView.extend({
         if (initialLayer) {
             map.addLayer(initialLayer);
         }
-
+        */
         map.addLayer(this._areaOfInterestLayer);
         map.addLayer(this._modificationsLayer);
-        */
```
lewfish: Consider getting rid of this commented out code.

lewfish:
You can merge it. Just check that the AOI is still there were clicking back from Analyze.

src/mmw/js/src/core/views.js
kellyi: Feel free to cut it if it's not necessary! Can't remember why I'd left it in the first place, but if we need it again, we can find it in the git history.

arottersman:
Cool, thanks! The AOI is still their when you leave the analyze view. I also a pushed a fixup that removes the commented code.

Issue #33, opened 2016-11-10
Missing Scenario Tab Dropdowns


rajadain:
Not sure how this happened, but all scenario tabs are missing the dropdown menus. There are corresponding failing tests for this. Get them back.

Issue #32, opened 2016-11-10
Back Button Broken in Modeling View


rajadain:
After a user has drawn a shape and entered the modeling view, if they click "Back" they are taken to the Draw view, but in a broken state. Make it so they are taken to a clean Draw view, with no area of interest and a clean slate to draw anew.

PR #29, opened 2016-11-09
Add Draw Sidebar

rajadain:
## Overview

Adds a sidebar in the "Draw" stage with the drawing tool.

### Demo

bee-draw-sidebar

### Notes

There's an upcoming PR that gets rid of the Analyze view. With that in place, the "Confirm" button will stop crashing.

I'm putting this up quickly so other people have an idea of what I'm doing so they can start working off of this.

## Testing Instructions

Pull down this branch, run the bundle script, and open [http://localhost:8000](http://localhost:8000). Ensure you are able to draw a shape, cancel drawing, and re-draw after finishing it.

Connects #22

Issue #23, opened 2016-11-07
Skip Analyze


rajadain:
According to the [wireframes](https://github.com/project-icp/bee-pollinator-app/files/576074/2016_10_27_bee-pollinator_v1.pdf), we go directly from "Draw" to "Modeling". There is no intermediary "Analyze" step. Remove the front-end components that dealt with that, and wire up the transition to go directly from "Draw" to "Modeling".

The "Modeling" stage can have non-functional / placeholder charts for now.

Depends on #22

PR #34, opened 2016-11-10
Skip Analyze

rajadain:
## Overview

After a shape is drawn, and the user clicks "Confirm", they are taken to the modeling view instead of analyze. Analyze is completely removed from the front-end: it no longer exists as its own page or as a tab in the modeling view. The back button in the modeling view has also been removed.

### Demo

bee-transition-to-model

### Notes

The modeling view currently fires TR-55, which was easier than setting up a dummy endpoint.

I've created #32 and #33 for issues discovered, but not addressed, while working on this.

## Testing Instructions

Pull down the branch. Run the bundle script:

```
./scripts/bundle.sh --debug --vendor
```

Load [http://localhost:8000](http://localhost:8000). Draw a shape, and click "Confirm". You should be taken to the modeling view with the correct shape loaded. The actual model may not run (for uninvestigated reasons).

Builds on top of #29 
Connects #23


arottersman:
Tested, and looks good!

Consider also  removing:
- analyze api endpoints https://github.com/project-icp/bee-pollinator-app/blob/54aa0ca7a93a6e0841fa397646ede7a624ea1a58/src/icp/apps/home/urls.py#L23
- `src/icp/sass/pages/_analyze.scss`

rajadain:
Good point, also removed in the latest push.

rajadain:
Also rebased to remove merge conflicts.

arottersman:
+1!

Issue #26, opened 2016-11-09
Add Basemaps


rajadain:
Add a satellite and street basemap, and a layer toggle button to switch between them. Refer to wireframes for details.

Issue #21, opened 2016-11-07
Update map to match wireframe


rajadain:
- [ ] The draw outline color should be the same `$brand-primary` yellow
 - [ ] The zoom and full-screen map controls go to the bottom right and are now horizontally arranged
 - [ ] Intstead of a complex layer selector, there are two simple buttons to switch between? or toggle? layers
 - [ ] Search remains on the top left, but all other "Draw" and "Reset" buttons are removed

Example:

![image](https://cloud.githubusercontent.com/assets/1430060/20072121/3015e14c-a4f5-11e6-906f-92e4085fd60e.png)

Full wireframes: https://github.com/project-icp/bee-pollinator-app/files/576074/2016_10_27_bee-pollinator_v1.pdf

ajrobbins:
>there are two simple buttons two switch between? or toggle? layers
@jfrankl can you clarify?

jtarricone:
Does the app start in "draw" mode, given that the draw controls have been removed in the design? If so, should that behavior be implemented with the rest of these changes?

jtarricone:
@jfrankl is there a stylesheet or reference for the color palette?

rajadain:
From @jfrankl:

 * The first, bottom-left button will toggle between Satellite and Streets.
 * The second button will toggle an overlay layer

Also, the color palette should be updated as part of #20. @arottersman had the idea of color dropping from the wireframes, so once those are in the drawing color here should update automatically.

Issue #20, opened 2016-11-07
Update name, fonts and branding colors


rajadain:
- [ ] App header should say "Bee Pollinator Decision Support Tool"
 - [ ] `$brand-primary` should be a particular shade of yellow rather than teal
 - [ ] Heading don't look like Roboto anymore

Example:

![image](https://cloud.githubusercontent.com/assets/1430060/20071955/8e409a42-a4f4-11e6-8775-55338b921f1c.png)

Full wireframes available here: https://github.com/project-icp/bee-pollinator-app/files/576074/2016_10_27_bee-pollinator_v1.pdf

Consult with @jfrankl before proceeding.

PR #35, opened 2016-11-10
Feature/touchup map

jtarricone:
This adds basemaps to the UI and updates the leaflet controls to reflect the new design. I used placeholder icons for the controls, and I deferred changing stylesheets in favor of #20 , which might require rebasing of one of the branches.

To test:
* clone this branch and bring up the environement
* navigate to `http://localhost:8000`
* check that the leaflet map has a basemap
* check that the controls resemble the design here: https://github.com/project-icp/bee-pollinator-app/files/576074/2016_10_27_bee-pollinator_v1.pdf
* click on the basemap toggle control a few times (the second button in the lower left) and check that it switches the map
* do some panning and zooming and toggle at random
* check that that the existing controls behave as expected; e.g. the search control searches, the zoom buttons zoom
* check the browser console for any errors

Connects #21 Connects #26

### DIFF
--- a/src/icp/js/src/core/layerControl.js
+++ b/src/icp/js/src/core/layerControl.js
@@ -1,0 +1,318 @@
+'use strict';
+
+var L = require('leaflet'),
+    $ = require('jquery'),
+    _ = require('underscore'),
+    Marionette = require('../../shim/backbone.marionette'),
+    App = require('../app'),
+    layerControlButtonTmpl = require('./templates/layerToggleControl.html'),
+    layerControlListTmpl = require('./templates/layerControlListTmpl.html');
+
+module.exports = L.Control.Layers.extend({
+
+    // Copied from  https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js
+    // with modifications for allowing a third layer type (observations)
+    initialize: function (baseLayers, overlays, options) {
+        L.setOptions(this, options);
+
+        this._layers = {};
+        this._lastZIndex = 0;
+        this._handlingClick = false;
+
+        for (var i in baseLayers) {
+            this._addLayer(baseLayers[i], i);
+        }
+
+        for (i in overlays) {
+            this._addLayer(overlays[i], i, 'overlay');
+        }
+    },
+
+
+    // Somewhat copied from https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js,
+    // with modifications to use a customer container element and our own way to toggle visibility.
+    _initLayout: function() {
+        var className = 'leaflet-control-layers',
+            container = this._container = new LayerControlButtonView({
+                layerControl: this
+            }).render().el,
+            listContainer = this._listContainer = new LayerControlListView({}).render().el;
+
+        this._form = $(listContainer).find('form').get(0);
+
+        // Copied directly from the parent class.
+        // makes this work on IE touch devices by stopping it
+        // from firing a mouseout event when the touch is released
+        container.setAttribute('aria-haspopup', true);
+
+        // Copied directly from the parent class, with addition of listContainer
+        L.DomEvent.disableClickPropagation(container);
+        L.DomEvent.disableClickPropagation(listContainer);
+        if (!L.Browser.touch) {
+            L.DomEvent.disableScrollPropagation(container);
+            L.DomEvent.disableScrollPropagation(listContainer);
+        }
+
+        // Expand the layer control so that Leaflet
+        // knows it's shown. We don't toggle this, but
+        // instead we toggle visibility using our own
+        // methods.
+        //this._expand();
+
+        // Kept around so we don't have to override _update, even though
+        // it's not used in the UI.
+        this._separator = L.DomUtil.create('div', className + '-separator');
+
+        // Add container element for each layer type
+        this._baseLayersList = L.DomUtil.create('div', className + '-base',
+                $(listContainer).find("#basemap-layer-list").get(0));
+        this._overlaysList = L.DomUtil.create('div', className + '-overlay',
+                $(listContainer).find("#overlays-layer-list").get(0));
+
+        // Add the layer control list as a sibling to the map.
+        // This lets us control the size of the list based
+        // on the height of the map.
+            //$('#map').append(listContainer);
+    },
+
+    // Overrides the parent class and allows separation of overlay layers
+    // by type.
+    _addItem: function (obj) {
+        var label = document.createElement('label'),
+            checked = this._map.hasLayer(obj.layer),
+            input,
+            controlName = 'leaflet-overlay-layers';
+
+        if (obj.overlay && obj.overlayType === 'observation') {
+            controlName = 'leaflet-control-layers-observation';
+            input = document.createElement('input');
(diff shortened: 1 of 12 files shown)


jtarricone:
After rebasing: it looks like the sidebar toggle button will have to be moved, it's stacked atop the zoom controls. I'll push a fixup shortly.

jtarricone:
I pushed a fix for the sidebar toggle, take a look when you get a chance.

rajadain:
Taking a look now

src/icp/sass/base/_map.scss
```diff
@@ -135,4 +136,54 @@
...
+  display: flex;
+  flex-direction: row-reverse;
+}
+
+.leaflet-bottom.leaflet-left {
+  display: flex;
+  flex-direction: row;
+}
+
+.leaflet-control-zoom-in {
```
rajadain: Rather than removing each corner radius on the zoom tools individually, prefer the simpler removal from all Leaflet controls as indicated in the wireframe:

```diff
diff --git a/src/icp/sass/base/_map.scss b/src/icp/sass/base/_map.scss
index 47e4552..9b754c9 100644
--- a/src/icp/sass/base/_map.scss
+++ b/src/icp/sass/base/_map.scss
@@ -174,16 +174,6 @@
   flex-direction: row;
 }
 
-.leaflet-control-zoom-in {
-  border-top-left-radius: 0px !important;
-  border-top-right-radius: 0px !important;
-  border-bottom-left-radius: 0px !important;
-  border-bottom-right-radius: 0px !important;
-}
-
-.leaflet-control-zoom-out {
-  border-top-left-radius: 0px !important;
-  border-top-right-radius: 0px !important;
-  border-bottom-left-radius: 0px !important;
-  border-bottom-right-radius: 0px !important;
+.leaflet-control a {
+    border-radius: 0 !important;
 }
```

src/icp/js/src/core/views.js
```diff
@@ -257,6 +256,11 @@ var MapView = Marionette.ItemView.extend({
         map.addLayer(this._modificationsLayer);
 
         map.addControl(new SidebarToggleControl());
+
+        // wrap the zoom & sidebar toggle controls in one div for styling
+        $('.leaflet-bottom.leaflet-right>.leaflet-control-sidebar-toggle, \
+          .leaflet-bottom.leaflet-right>.leaflet-control-zoom')
+            .wrapAll('<div style="display:flex;flex-direction:row;float:right;"></div>');
```
rajadain: I see why this is required, but it would be preferable to wrap this in a div with a class, and apply the style in `_map.scss`.

jtarricone:
Whoooa what is up with the two extra commits? Will that resolve itself upon merging or should I try to figure out what happened?

rajadain:
I merged that on top of Alice's work in yesterday. If you rebase your commits over `develop` it should be fine. Those first two commits have already been squashed and are part of develop, and can be safely dropped from this branch.

src/icp/sass/base/_map.scss
```diff
@@ -135,4 +136,54 @@
...
+  border-top-right-radius: 0px !important;
+  border-bottom-left-radius: 0px !important;
+  border-bottom-right-radius: 0px !important;
+}
+
+.leaflet-control-zoom-out {
+  border-top-left-radius: 0px !important;
+  border-top-right-radius: 0px !important;
+  border-bottom-left-radius: 0px !important;
+  border-bottom-right-radius: 0px !important;
```
rajadain: Also consider adding something like this:

```diff
diff --git a/src/icp/sass/base/_map.scss b/src/icp/sass/base/_map.scss
index 47e4552..7ebc88c 100644
--- a/src/icp/sass/base/_map.scss
+++ b/src/icp/sass/base/_map.scss
@@ -187,3 +187,12 @@
   border-bottom-left-radius: 0px !important;
   border-bottom-right-radius: 0px !important;
 }
+
+.leaflet-bar a, .leaflet-bar a:hover {
+    border-bottom: none;
+    border-right: 1px solid #ccc;
+}
+
+.leaflet-bar a:first-child {
+    border-right: none;
+}
```

to fix the odd borders caused by changed orientation of the zoom control.

src/icp/js/src/core/views.js
rajadain: Also, change the `flex-direction` to `row-reverse` to match the wireframes.

rajadain:
I just merged in my other PR. You can rebase on top of develop to get the latest. Thankfully it doesn't look like there's any merge conflicts.

src/icp/sass/base/_map.scss
```diff
@@ -136,34 +136,32 @@
...
   }
 }
 
 .regular-text {
-  color: black;
-  text-decoration: none;
   align-self: center;
 }
 
+.wrapped-content {
```
rajadain: This may be better named `.leaflet-bottom-right-controls` for explicitness.

src/icp/js/src/core/layerControl.js
```diff
@@ -0,0 +1,318 @@
...
+
+var L = require('leaflet'),
+    $ = require('jquery'),
+    _ = require('underscore'),
+    Marionette = require('../../shim/backbone.marionette'),
+    App = require('../app'),
+    layerControlButtonTmpl = require('./templates/layerToggleControl.html'),
+    layerControlListTmpl = require('./templates/layerControlListTmpl.html');
+
+module.exports = L.Control.Layers.extend({
```
mmcfarland: Is it still necessary to extend this Layers control?  This has a bunch of custom logic for creating tabs and custom behavior for the different layer types of MMW.  In this much simplified toggle, you may be able to use the standard `L.Control.Layers` - or none at all?

src/icp/js/src/core/layerControl.js
```diff
@@ -0,0 +1,318 @@
...
+    },
+
+    events: {
+        'click @ui.controlToggle': 'toggle'
+    },
+
+    toggle: function() {
+        var map = this.layerControl._map;
+        if (this.hasBeenToggled) {
+            var newLayer = _.find(this.layerControl._layers, function(layer) {
```
mmcfarland: Instead of using the layer control to search for layers, we could probably just use the layer configuration directly.  It would be simpler as well once the extraneous layers are removed.  I don't see the layer control being used for anything other than a container of layers which are added/removed directly here.

rajadain:
+1 tested, works well, looks good! I did see a "Locate Me" button in the top right that wasn't in the wireframes

image

That wasn't working for me, so maybe we can take that out? We can also leave it in.

Adding back the heavier Layer Selector stuff does incur some technical debt as we'll have to go back and simplify it, as @mmcfarland suggests above. It would be nice to address it here, given the leanness of this project. But we may also want this in for today's demo with @ajrobbins. They may be able to speak more to which is preferable.

Also heads up that there's merge conflicts.

mmcfarland:
Yeah, that's fair enough.  Let's move ahead with this and iterate later.  I have a good deal of experience in the layer selector and could probably trim it up rather quickly in the future.

rajadain:
+1 the locate control is gone. I'm going to make an issue to clean up the Layer Selector. This is good to go!

Issue #30, opened 2016-11-09
Add Numeric Text Input for Hives


rajadain:
Instead of the precipitation slider, the Bee Polliinator model takes a numeric input for the number of managed honeybee hives per unit area. Remove the slider and add a text input in the Modeling view.

PR #36, opened 2016-11-11
Replace Precipitation Slider With Number Of Hives Input

arottersman:
## Overview
Bee pollinator model takes the number of managed hives as a numeric input, so this PR replaces the precipitation slider an number input field.

### Other small fixups
- Match the inverted dropdown button "Modifications" in the model header to the wireframes (ie make it blue)
- Set "New Scenario" as initial active tab

screen shot 2016-11-11 at 3 15 11 pm
screen shot 2016-11-11 at 3 15 22 pm

## Testing Instructions
- Pull this branch
- `./scripts/bundle.sh`
- Draw a shape
- You should only be able to enter numeric values in the input field
- You should be able go to compare still
- There shouldn't be any errors in the console

Connects #30

### DIFF
--- a/src/icp/js/src/modeling/controls.js
+++ b/src/icp/js/src/modeling/controls.js
@@ -14,7 +14,6 @@ var $ = require('jquery'),
     models = require('./models'),
     modificationConfigUtils = require('./modificationConfigUtils'),
     gwlfeConfig = require('./gwlfeModificationConfig'),
-    precipitationTmpl = require('./templates/controls/precipitation.html'),
     manualEntryTmpl = require('./templates/controls/manualEntry.html'),
     userInputTmpl = require('./templates/controls/userInput.html'),
     inputInfoTmpl = require('./templates/controls/inputInfo.html'),
@@ -35,7 +34,8 @@ var ControlView = Marionette.LayoutView.extend({
         this.mergeOptions(options, [
             'controlModel',
             'addModification',
-            'addOrReplaceInput'
+            'addOrReplaceInput',
+            'numberofhives'
         ]);
     },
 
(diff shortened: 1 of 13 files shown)


rajadain:
I see a gray box around the modification dropdown when there are 0 modifications:

image

but not when there's 1 or more:

image

Can we stick to one or the other? I think the gray box is good, it makes the text more readable and the clickability more obvious.

rajadain:
I found that switching between scenarios is broken. Created #42 to fix that. Also found the tiny maps in the Compare view have an "Expand" Leaflet control. Created #44 to get rid of those.

Other than that, this works well. Going to take a quick peek at the code.

src/icp/js/src/modeling/controls.js
```diff
@@ -415,139 +415,32 @@ var ConservationPracticeView = ModificationsView.extend({
...
-    },
-
-    onSliderChanged: function() {
-        var value = parseFloat(this.ui.slider.val()),
-            // Model expects Imperial inputs.
-            imperialValue = coreUtils.convertToImperial(value, 'cm'),
-            modification = new models.ModificationModel({
+    onInputChange: function() {
+        var value = parseInt(this.ui.input.val());
+        var modification = new models.ModificationModel({
```
rajadain: When declaring a number of variables, prefer to do them in one statement:

```js
var value = parseInt(this.ui.input.val()),
    modification = new models.ModificationModel({
        ...
    });
```

in keeping with the project's style.

arottersman:
The gray box around modifications is actually the disabled styling coming in from bootstrap, and it appears on 0 Modifications for OG MMW as well. I pushed a fixup that gets rid of it so that we can match the wireframes, but I agree it looks more clickable than the plain text button...

src/icp/js/src/modeling/templates/controls/userInput.html
```diff
@@ -1,5 +1,5 @@
-<div class="form-group">
-  <label for="{{ userInputName }}">{{ displayNames[userInputName] }}</label>
-  <input type="text" class="form-control" id="{{ userInputName }}">
+<div class="hives">
```
rajadain: `.form-group` → `.hives` ?

arottersman: This was `form-group` because it was in one of those gwlfe dropdowns (should we get rid of those?)
screen shot 2016-11-14 at 12 12 06 pm
I've set it to `hives` to follow how `precipitation` had its own class for styling. Should I rename it?

rajadain: Ah I see. No, we can keep it as `.hives` then. I thought it strange since everywhere else `.precipitation` was replaced with `.hives`.

rajadain:
Friendly reminder to squash before merging.

src/icp/js/src/modeling/models.js
```diff
@@ -1008,13 +1008,13 @@ function getControlsForModelPackage(modelPackageName, options) {
         if (options && (options.compareMode ||
                         options.is_current_conditions)) {
             return new ModelPackageControlsCollection([
-                new ModelPackageControlModel({ name: 'precipitation' })
+                new ModelPackageControlModel({ name: 'landcover' })
```
arottersman: From @ajrobbins  -- Land Cover modifications should appear even on current conditions tab

arottersman:
Thanks @rajadain for reviewing!

Issue #39, opened 2016-11-11
Remove expand/collapse from compare view map images


ajrobbins:
![screen shot 2016-11-11 at 4 25 39 pm](https://cloud.githubusercontent.com/assets/12868342/20231012/ef120186-a82c-11e6-99cf-dc97a099640d.png)

Should look like it does on MMW:

![screen shot 2016-11-11 at 4 36 14 pm](https://cloud.githubusercontent.com/assets/12868342/20231026/02a6010c-a82d-11e6-819b-45883054d17c.png)

arottersman:
Shares a point with https://github.com/project-icp/bee-pollinator-app/pull/41

PR #41, opened 2016-11-14
Fix Scenario Dropdowns and Compare View Map Sidebar Toggle

arottersman:
### Overview
During #12 we removed the template helpers of `ScenarioTabPanelView` - they set booleans we needed for the dropdowns on that view to appear. This PR adds the necessary template helpers back in.

Because it was such a quick fix and I had noticed what was causing the expand/collapse control to appear in the compare view (#39) I've also included a0feb80.

_Dropdown is back_
screen shot 2016-11-14 at 10 15 21 am
screen shot 2016-11-14 at 10 15 29 am

_Sidebar toggle no longer appears on compare maps_
screen shot 2016-11-14 at 10 05 28 am

### Testing Instructions
- Pull this branch
- `./scripts/bundle.sh`
- Make sure the sidebar expand/collapse toggle still appears and works in the draw view
- Go to the model view and confirm the scenario dropdowns are there/working (they will not appear for "Current Conditions"
- Go to compare view and confirm there are no sidebar toggles on the little maps

Connects #33, #39

### DIFF
--- a/src/icp/icp/settings/base.py
+++ b/src/icp/icp/settings/base.py
@@ -470,6 +470,7 @@ MAP_CONTROLS = [
     'LayerAttribution',
     'LocateMeButton',
     'ZoomControl',
+    'SidebarToggle'
 ]
 
 TR55_PACKAGE = 'tr-55'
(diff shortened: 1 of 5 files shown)


jtarricone:
+1 Tested

arottersman:
Thanks, Joe!

PR #48, opened 2016-11-16
Make dev environment accessible from internal n/w

rajadain:
Same as https://github.com/azavea/pwd-stormdrain-marking/pull/268. You know the drill.


arottersman:
I think we'll also need to add `localhost`

rajadain:
Good call. Added.

arottersman:
+1, tested!

PR #54, opened 2016-11-18
Fix UI error when transitioning from model view

jtarricone:
This fixes an error that resulted in the UI being put into a bad state when app was backed out of the modelling view.

To test:
* clone this branch and bring up the environment
* navigate to `http://localhost:8000`
* draw an AoI and confirm it to move to the modelling view -->
download 7
* back out of this using the browser, and you should be returned to the same home view -->
download 8
* check the browser console for any errors; the `Failed to start modeling job.` and associated 500 error are expected

Connects #32

### DIFF
--- a/src/icp/js/src/core/views.js
+++ b/src/icp/js/src/core/views.js
@@ -181,6 +181,11 @@ var MapView = Marionette.ItemView.extend({
     // Google Maps API library is loaded asynchronously via an inline script tag
     _googleMaps: (window.google ? window.google.maps : null),
 
+    _centeredBounds: [
+        [24.2, -126.4],
+        [49.8, -66.0]
+    ],
+
     initialize: function(options) {
         var defaultLayer = _.findWhere(settings.get('base_layers'), function(layer) {
                 return layer.default === true;
@@ -205,10 +210,7 @@ var MapView = Marionette.ItemView.extend({
             });
 
         // Center the map on the U.S.
-        map.fitBounds([
-            [24.2, -126.4],
-            [49.8, -66.0]
-        ]);
+        map.fitBounds(this._centeredBounds);
 
         this._leafletMap = map;
         this._areaOfInterestLayer = new L.FeatureGroup();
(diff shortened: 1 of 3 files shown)


src/icp/js/src/modeling/controllers.js
```diff
@@ -169,7 +169,7 @@ function setPageTitle() {
         modelPackageDisplayName = _.find(modelPackages, {name: modelPackageName}).display_name;
 }
 
-function projectCleanUp() {
+function projectCleanUp(clearMapState) {
```
jtarricone: I made the map state clear behavior parameterized because I don't think we yet know whether we'll want to support the old MMW behavior of stashing the AoI and associated data if the user discards a model.

src/icp/js/src/core/views.js
```diff
@@ -266,6 +266,18 @@ var MapView = Marionette.ItemView.extend({
             .wrapAll('<div class="leaflet-bottom-right-controls"></div>');
     },
 
+    clearMapState: function() {
+        var self = this;
+        self._leafletMap.fitBounds([
```
jtarricone: These bounds are leftovers from the MMW code; it actually looks like the post-cleared map is zoomed further out than the initial one, so these may need to be updated.

arottersman: Not sure why it's zooming, but consider keeping the bounds somewhere in the settings - that way we don't have to maintain them both here and [here](https://github.com/jtarricone/bee-pollinator-app/blob/73cd74cda08868cab87fa45947f5ddec42372567/src/icp/js/src/core/views.js#L208)

src/icp/js/src/modeling/controllers.js
```diff
@@ -184,9 +184,14 @@ function projectCleanUp() {
         App.projectNumber = scenarios.at(0).get('project');
     }
 
-    App.getMapView().updateModifications(null);
     App.rootView.subHeaderRegion.empty();
     App.rootView.sidebarRegion.empty();
+
+    // when invoked as part of a jquery chain, clearMapState will be an instance
```
arottersman: When does a jquery chain invoke this method? Also, consider renaming `clearMapState` to something like `shouldClearMapState` to clarify it should be a boolean.

jtarricone: Good catch, that was how it was invoked from the analysis view, which I had to address when I initially implemented this but which is no longer the case since that view was removed.

src/icp/js/src/core/views.js
jtarricone: I pushed an update that declares the bounds as a class-level member of the view, to reduce the redundancy.

arottersman:
+1!

Issue #46, opened 2016-11-16
Clean Up Layer Selector Code


rajadain:
Following @mmcfarland's comments on #35, clean up the layer selector code to

 * Remove `get_layer_url` method which is no longer needed since there are no custom layers
 * Remove all extraneous blocks in `layer_settings.py`
 * Don't extend the `L.Control.Layers` since we don't need the complex tab structure anymore
 * Use the layer configuration directly in the layer control, instead of `_.find`ing them

PR #55, opened 2016-11-18
Simplify layer selector

mmcfarland:
The layer control for this app is a simple toggle button so the more
complex arrangement for varying layer types has been removed and a more
direct method of adding the toggle to the map replaces it.

I also swapped out the old streets layer for a free, temporary ESRI one since the previous mapbox was tied to an account owned and paid for by a different client.  This isn't necessarily meant to be the one going forward.

Connects #46 

#### Testing
* ./scripts/bundle.sh --debug
* Layers should toggle as before


src/icp/js/src/core/layerControl.js
```diff
@@ -247,8 +30,9 @@ var LayerControlButtonView = Marionette.ItemView.extend({
     className: 'leaflet-control leaflet-control-layers',
 
     initialize: function(options) {
-        this.layerControl = options.layerControl;
-        this.hasBeenToggled = false;
+        this.layers = options.layers;
+        this.map = options.map;
+        this.satelliteVisible = true;
```
arottersman: Consider passing the initial layer in the `options` dictionary, then setting this boolean based on whether it's satellite or not. That way if want to change the initial layer to streets we'll only have to change it in the `MapView`

arottersman:
+1, tested - toggle still works as expected!

src/icp/js/src/core/layerControl.js
mmcfarland: Good idea, will do.

Issue #47, opened 2016-11-16
Add panel for landing view


ajrobbins:
From wireframes:

![screen shot 2016-11-16 at 4 40 02 pm](https://cloud.githubusercontent.com/assets/12868342/20366946/6a93af98-ac1b-11e6-81ff-839c01c79f96.png)

PR #56, opened 2016-11-21
Add Splash Page

rajadain:
## Overview

Adds a splash view for users to land on, so they can get some context about the application. Once ready, they can click "Get started" to move to the drawing area.

### Demo

image

## Testing Instructions

 * Check out this branch
 * Bundle vendor scripts to include new image:

        ./scripts/bundle.sh --vendor

 * Open the home page [localhost:8000](http://localhost:8000/). You should see the splash screen.
 * Click "Get started". You should be taken to the draw view.
 * Drawing should work as expected.
 * Back / forward buttons should work as expected.
 * You should be able to go directly to the draw view at [localhost:8000/draw](http://localhost:8000/draw)

Connects #47


ajrobbins:
Has the geolocator been removed from this page?

rajadain:
The geolocator search is only available in the draw view, but taking another look at the wireframes I see it should be available in the splash screen as well. Fixing now.

rajadain:
Done:

image

rajadain:
Thanks for taking a look. Squashed fixups. Merging now ...

Issue #64, opened 2016-12-02
Small demo feedback


mmcfarland:
In preparing for the demo for the ICP annual meeting, I made some small changes:
* allow login
* show land cover overlay (and cleanup)
* update label for managed hives

PR #65, opened 2016-12-02
Small demo improvements

mmcfarland:
#### Overview
Some miscellaneous fixes that added small polish to a recent demo:
* Fix syntax error for django auth backend which prevented user from logging in
* Show an overlay layer from Crop (using NLCD)
    * Rename code and files to more clearly represent that the control toggles an overlay, not an opacity slider.
    * Use a temporary overlay (NLCD) while the crop layer is being developed
    * Update the code to remove redundancies, read from configuration and have updated text
* Update text label for Managed Hives control from `Value` to `Managed Hives per Acre`

Connects #64 

#### Testing
1. Use the crop data toggle button
    * Should show the NLCD
    * Updated text to "Crop Layer" from "CDL Layer"
    * Added tooltip which explains the action clicking will toggle (show/hide)
2. Create a user `./scripts/manage.sh createsuperuser`
    * Log in to the app using the above credentials
3. Verify text for "Value" has been replaced with more meaningful text

imageedit_4_7795560672


src/icp/js/src/core/overlayControl.js
```diff
@@ -0,0 +1,58 @@
...
+        } else {
+            this.map.addLayer(this.layer);
+        }
+        this.isDisplayed = !this.isDisplayed;
+        this.render();
+    },
+
+    templateHelpers: function() {
+        return {
+            iconClass: this.hasBeenToggled ? "fa fa-eye-slash" : "fa fa-eye",
```
arottersman: I think we need to update this to `!this.isDisplayed` . Currently the eye doesn't close on click

mmcfarland: Thanks, I think this got dropped when I rebased and had a conflict with your update.

mmcfarland:
Updated with fb7b0b5 to address bad conflict resolution.

Issue #53, opened 2016-11-17
Provision worker with model dependencies


mmcfarland:
Provision the worker vm with ansible roles to support the model dependencies:
* GDAL
* TBD

caseypt:
Blocked until other model dependencies are figured out.

ajrobbins:
@mmcfarland is there an update on what the other model dependencies are?

mmcfarland:
Yes, though this work is actually being done in tandem with https://github.com/project-icp/bee-pollinator-model/issues/3.  I think it was split into two cards for purpose of Waffle points.  I've assigned and moved it, sorry for the confusion.

PR #68, opened 2016-12-05
Integrate initial pollinator model

mmcfarland:
#### Overview
Adds an initial version of the pollinator model to the app and makes it callable by the front end.  It also removes a considerable amount of modelling/geoprocessing code that was needed for MMW but is not for this project.  I decided to create the model as a python module, but one that was embedded in this repo.  I'm hoping that this will reduce the feedback cycle during development as opposed to having it in its own repo with a deployment process.  The module could be plucked out in the future when we want a properly versioned and separate product.  

#### Notes
* This is a big PR, with a lot of files getting touched.  My recommendation for review is:
    * The 3 commits are pretty well isolated, I think it makes sense to evaluate them in sequence
    * The majority of the changes are wholesale deletions
    * There is a chunk of code for adding a python `pollinator` module.  This is fairly initial, and I wouldn't recommend too much time reviewing the _implementation_ of that code, but more perhaps the structure and how it is included in the overall project.  I anticipate near term work to copy over some tests add comments etc, which will probably be a better time to review and evaluate the implementation
* The model is presently returning a fairly meaningless number, but it's derived from an interim stage of the model (farm abundance).
* I tried to strike a balance between removing unneeded code and just letting the existing (and quicker to implement) code from MMW function.  For example, there is only one model "package" now, but I left a lot of code that assumed there may be multiple.

Connects #53 
Connects https://github.com/project-icp/bee-pollinator-model/issues/3

#### Testing
* Get the raster data from @mmcfarland
* Bring up your vagrant machines
* Ensure that you have created the initial model tables (`./scripts/manage.sh migrate`)
* `vagrant provision worker`
* Choose a small AoI (< 400 acres) and create a model project
* In DevTools verify that the initial `yield` endpoint is requested and that after polling, a numerical result is returned.  This applies to each scenario and after modifications are made


src/icp/apps/modeling/tasks.py
```diff
@@ -4,308 +4,19 @@
...
-        value = modification['value']
-
-        if name == 'landcover':
-            return {'change': ':%s:' % value['reclass']}
-        elif name == 'conservation_practice':
-            return {'change': '::%s' % value['bmp']}
-        elif name == 'both':
-            return {'change': ':%s:%s' % (value['reclass'], value['bmp'])}
+def calculuate_yield(model_input):
+    from pollinator import crop_yield
```
mmcfarland: Not sure if this is a good solution.  The problem is that the module is _parsed_ on `app` when Django loads the celery tasks, but the task is _executed_ on `worker`.  Since `app` isn't provisioned with these dependencies, it fails if the modules are loaded at the top.  Importing them within the task function means they're only loaded on `worker`.

src/icp/apps/modeling/tasks.py
```diff
@@ -4,308 +4,19 @@
...
-        name = modification['name']
-        value = modification['value']
-
-        if name == 'landcover':
-            return {'change': ':%s:' % value['reclass']}
-        elif name == 'conservation_practice':
-            return {'change': '::%s' % value['bmp']}
-        elif name == 'both':
-            return {'change': ':%s:%s' % (value['reclass'], value['bmp'])}
+def calculuate_yield(model_input):
```
rajadain: Typo here `calculuate` → `calculate`

src/icp/js/src/modeling/models.js
```diff
@@ -734,10 +612,9 @@ var ScenarioModel = Backbone.Model.extend({
             });
 
         if (needsResults && self.fetchResultsPromise === undefined) {
-            var fetchResults = _.bind(self.fetchResults, self),
-                fetchGisDataPromise = App.currentProject.fetchGisDataIfNeeded();
+            var fetchResults = _.bind(self.fetchResults, self);
 
-            self.fetchResultsPromise = fetchGisDataPromise.then(function() {
+            self.fetchResultsPromise = $.when().then(function() {
```
rajadain: Can we remove the `$.when()` entirely? And just bring the inner function out of the callback?

rajadain:
I had to

```
./scripts/manage.sh makemigrations
./scripts/manage.sh migrate
```

to incorporate the changes to the `model_package` field, maybe because my database was already provisioned on `develop`? Consider adding it to this PR:

```python
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('modeling', '0019_project_gis_data'),
    ]

    operations = [
        migrations.AlterField(
            model_name='project',
            name='model_package',
            field=models.CharField(help_text='Which model pack was chosen for this project', max_length=255, choices=[('yield', 'Crop Yield Model')]),
        ),
    ]
```

src/icp/apps/modeling/tasks.py
rajadain: I was going to suggest moving the contents of this function to another file, which gets called inside this task, but seeing how small this file already is that may be less efficient. However, we have had separation between functionality and Celery Task glue before, so perhaps that's not a bad idea.

rajadain:
Will take another look tomorrow morning, but so far this looks really good! When I tested the first time, I got back a `results: "NaN"` for one scenario and a real number for the other, but every test since has resulted in the same number for both scenarios (since there is no difference between them I expect them to be identical), so it must have been a hiccup of some sort.

This was the input:

```json
{
    "aoi_census": null,
    "area_of_interest": {
        "coordinates": [
            [
                [
                    [
                        -75.16597867012023,
                        39.95458120075746
                    ],
                    [
                        -75.16676187515259,
                        39.95286231520063
                    ],
                    [
                        -75.16339302062988,
                        39.95309259902528
                    ],
                    [
                        -75.16597867012023,
                        39.95458120075746
                    ]
                ]
            ]
        ],
        "type": "MultiPolygon"
    },
    "inputmod_hash": "260b0512813a3b96975913f1fe7993e6d751713988987e9331980363e24189ce",
    "inputs": [
        {
            "area": "0",
            "effectiveArea": null,
            "effectiveShape": null,
            "effectiveUnits": null,
            "name": "numberofhives",
            "shape": null,
            "type": "",
            "units": "m<sup>2</sup>",
            "value": 0
        }
    ],
    "modification_censuses": null,
    "modification_hash": "d751713988987e9331980363e24189ce",
    "modification_pieces": []
}
```

and these the outputs:

```json
{
    "error": "",
    "finished": "2016-12-05T23:33:06.702Z",
    "job_uuid": "f28d15d9-440f-4dcb-9130-a3487ba8a903",
    "result": "NaN",
    "started": "2016-12-05T23:33:05.247Z",
    "status": "complete"
}
```

```json
{
    "error": "",
    "finished": "2016-12-05T23:33:08.586Z",
    "job_uuid": "0040325e-1503-4dc9-8b22-fbfbc9b8f499",
    "result": "2.4811235102334803e+25",
    "started": "2016-12-05T23:33:07.322Z",
    "status": "complete"
}
```

src/icp/apps/modeling/tasks.py
mmcfarland: I think we'd hit the same issue, if I understand the proposal.  If the new file was imported in `tasks.py`, it would still get parsed and its imports would raise.  I can't see a way around it other than tying the imports to the execution of the task function.  I'm also not sure if it's necessary hygiene since this is the only function in the file.

src/icp/js/src/modeling/models.js
mmcfarland: I initially tried that and couldn't make it work, I think because the presence of a value is also meaningful to some control flow code.  I'll take another pass, but if it's going to result in more than modest updates to the logic, I may just keep as-is.

src/icp/apps/modeling/tasks.py
rajadain: Hmm in that case I think it's fine. Even in the Scala geoprocessing code there are occasional function-scope imports, to account for implicits in the right place. This is no less hygienic than what we've done in the past.

mmcfarland:
Fixup: 30d7af8 for typo and migration - good catch.

I opted not to replace the promise, as it's exported and used around the app.

Thanks for the thorough look.

Issue #73, opened 2016-12-09
Change units from metric to imperial


ajrobbins:
We are currently using meters and km throughout the app. Convert to feet and acres.

mmcfarland:
Shares a point w/ #72

Issue #72, opened 2016-12-09
Text change


ajrobbins:
In search bar: `Search` --> `Enter address or location`

ajrobbins:
Model input: `Managed Hives` -->`Honeybee Hives Per Acre`

PR #83, opened 2016-12-19
Display Imperial Units instead of Metric

rajadain:
## Overview

Switches displayed values to imperial rather than metric. I could not find any examples of length, thus didn't convert meters to feet. More values will come in to play as we flesh out the model side, so we'll have to keep this in mind.

## Testing Instructions

Check out this branch, bundle scripts and hard refresh your browser.

 * Notice the new message in the search bar in the home / draw views.
 * Draw a very large area of interest (like half the country). The error message admonishing you should be in acres, not km².
 * Draw a regular area of interest and move to the model view. Draw a modification for New Scenario. Click the modification. Its area should be displayed in acres.
 * Click the modification dropdown in the top right of the model view. It should display the area of all drawn modifications in acres.

Are there any other places that may have metric units? If so, please let me know and I will update them.

Connects #72 
Connects #73

### DIFF
--- a/src/icp/js/src/core/models.js
+++ b/src/icp/js/src/core/models.js
@@ -225,13 +225,11 @@ var TaskMessageViewModel = Backbone.Model.extend({
 });
 
 var GeoModel = Backbone.Model.extend({
-    M_IN_KM: 1000000,
-
     defaults: {
         name: '',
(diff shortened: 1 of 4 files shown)


arottersman:
Reviewing this now!

Issue #81, opened 2016-12-15
Reclassify CDL into aggregate crop type group


mmcfarland:
Reclassify the CDL raster from individual crops to group types, defined by the `Attributes` column from https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/pollinator/src/pollinator/data/cdl_data.csv

* Provide new raster as 512x512 striped, compressed with LZW
* Generate a new CSV with new CDL type id, group and the remaining attributes (HN, HF)
* Try to retain dominant colormap for legend of new group id

PR #85, opened 2016-12-22
Aggregate CDL values by category

mmcfarland:
#### Overview
Two sets of crop data were provided to us by the ICP Team: cdl_data.csv containing mean scores for forage and nesting suitability by crop type and the CDL raster of land use cover describing crop distribution spatially.  The CDL has over 100 distinct types, but they were grouped into approximately 46 categories.  Two scripts were created which aggregate both data sets on that grouping to produce a more concise data set.  Some output of the scripts:
* A mapping of a new id to a list of old ids 
* A CSV of cdl_data compressed into group values only
* An uncompressed CDL tif with values reclassified according to that aggregation
* A compressed and tiled CDL tif suitable as a data source for the pollinator model

The model itself was updated to reference the new csv and tif as well as some assumptions about what crop types to report on based on the new ids.  The colormap for the tif was reduced as well, but the choices are temporary.

Connects #81 

#### Testing
* Reprovision the worker machine `vagrant provision worker`
* `vagrant ssh worker` and `cd /opt/app/pollinator/src/pollinator/reclass`
* Run `python group_cdl_data.py` to generate the new csv
* Run `python reclassify_cdl.py` to generate the two new tifs

The new CSV is committed in this PR but the compressed tif can be copied to your synced folder (`/opt/icp-crop-data` on your guest) at which point you can test the model by running the app and modeling a small AoI.
* Rebundle the javascript `./scripts/bundle.sh --debug`
* Search for `Drummond Dr, Yuba City, CA 95991` which should get you to an area with two pollinator dependent crop types: Almond (33) and Orchard (27)
screenshot from 2016-12-22 10 36 52

* Draw & confirm the output of the model conforms to the new grouped id scheme
```json
{
    "12": 0,
    "16": 0,
    "17": 0,
    "18": 0,
    "20": 0,
    "27": 0.8427295684814453,
    "33": 0.3804641366004944,
    "46": 0,
    "47": 0
}
```


src/icp/pollinator/src/pollinator/reclass/reclassify_cdl.py
```diff
@@ -0,0 +1,112 @@
...
+
+                    dst.write_band(1, reclass_data, window=window)
+
+                    if DEBUG:
+                        check_assignment(data, reclass_data, reclass_map)
+
+                    count += 1
+                    print_progress(count, total)
+
+                post_process()
```
arottersman: I was able to fix the following error by moving the call to `post_process` outside the file open scopes (ie just 1 indent in):
```
ERROR 1: Unable to compute a transformation between pixel/line
and georeferenced coordinates for ./cdl_reclass.tif.
There is no affine transformation and no GCPs.
```
With the error gdal would never create the second, compressed tif

mmcfarland: Will fix, thanks.

mmcfarland:
Added 7ae9489 to fixup up `post_propcess`. I believe `rasterio.open` establishes some GDAL global environment settings - this may have been interfering with our  additional gdal script.

src/icp/pollinator/src/pollinator/data/cdl_data.csv
```diff
@@ -62,7 +62,7 @@ CDL_code,CDL_categories,Attributes,Demand,HF.mean,HN.mean
 71,Other Tree Crops,Tree Crops,none,0.367659457,0.315017853
 72,Citrus,Citrus,little,0.358212334,0.28718199
 74,Pecans,Nuts,none,0.227475043,0.266643385
-75,Almonds,Almonds,great,0.344578281,0.324941214
+75,Almonds,Orchard,great,0.344578281,0.324941214
```
arottersman: Maybe this got messed up when I was destroying everything with git yesterday?

mmcfarland: Ah yep, that's it.  The checked in `cdl_data_grouped.csv` has the [correct value](https://github.com/project-icp/bee-pollinator-app/pull/85/commits/ceda5bcfc37dfd9c8aa908908e2c82b8b2d35e31#diff-a490a509170e66e9e165d0c3600fe9c6R34), but the source didn't keep it's update.  I think it was actually my fault when I reverted a change that altered the line endings in this file.

mmcfarland:
Added c4fbae8d02b3b0c466f04eadabd36a7a37739236 to prevent Almonds getting merged into Orchard.  I'm going to rerun the whole thing and make sure it still works.

Issue #84, opened 2016-12-20
Prevent new scenario tabs from executing bare model


mmcfarland:
The workflow for this app is:

1. Open current conditions first
2. Add land cover corrections to augment the base raster data layer
3. Go to new scenario to start adding enhancement or changes
  * Changes from 2) will be applied to 3) and subsequent scenarios
4. Add more scenario, repeat

The scenario tab should wait to execute until the tab is switched to.  If it is switched to and it has:
* no results yet (means hasn't executed yet) _and_ current conditions does have a result: don't execute the model, just get the results from current conditions, they will be the same
* no results yet and no results in CC, then we should execute as normal.

Let's chat about implementation details.

arottersman:
If landcover is added in a scenario should it be carried over to all subsequent scenarios?

mmcfarland:
No, unless that scenario is the current conditions.  I was assuming that would be a separate card because it involves re-executing all the scenarios as well as managing adding and removing those modifications.  I was picturing this card being a first step toward that ultimate goal.

PR #88, opened 2016-12-24
Use Current Condition Results In Scenarios Without Modifications

arottersman:
## Overview
Unlike MMW, land cover changes on the Current Conditions tab will carry over to the scenario tabs. Instead of running the model job for every open scenario, this PR sets up the model job to only kick off if: 
1) The scenario is current conditions (and it has no results/no resolved promise yet)
2) The scenario isn't current conditions, is active, and the current conditions job hasn't returned any results
3) There are new modifications

Otherwise we copy the current conditions results over to the initial "New Scenario", and don't rerun the model

## Demo
Below you can see we're on a scenario tab with scenario results, but only the current condition tab has called the model (there's only one set of polling calls)
screen shot 2016-12-27 at 12 52 34 pm

## Testing Instructions
- Pull this branch and `./scripts/bundle.sh --debug`
- Open the Dev tools console
- Draw an AOI and stay on the current conditions tab
   - Check that there is only one call to `/yield`
- When there are results, switch to the New Scenario tab
   - There should be no new polling, and the logged "Scenario Result" should be the same as Current Conditions' result
- Add some modifications in the new scenario and confirm the model re-runs
- Repeat the process, but while the Current Conditions tab is polling, switch to the New Scenario
   - The new scenario should start polling itself

Connects #84

### DIFF
--- a/src/icp/js/src/modeling/models.js
+++ b/src/icp/js/src/modeling/models.js
@@ -148,6 +148,12 @@ var ProjectModel = Backbone.Model.extend({
             promises.push(scenario.fetchResultsIfNeeded());
         });
 
+        this.get('scenarios').on('change:active', function(scenario) {
+            if (scenario.get('active')) {
+                scenario.fetchResultsIfNeeded();
+            }
+        });
+
         this.get('scenarios').on('add', function(scenario) {
             scenario.fetchResultsIfNeeded();
         });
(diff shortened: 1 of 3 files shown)


arottersman:
I was able to fix the re-polling by adding the `inputmod_hash` into the payload from the server. It was seeing it as `undefined` and thinking it always had stale results. I've also made the `always()` chaining conditional on there being `self.fetchResultsPromise`.

Not pushing a fixup yet because I noticed that 
* when you switch to a scenario tab while the current conditions is still polling, the current conditions polling stops
* when you go back to current conditions it doesn't re-poll
* if you add a modification to the scenario and then create a new scenario, the new scenario copies the old scenario's results

mmcfarland:
For the 3rd point, I believe there is a `resultModel` attribute for `polling` that you can check to see if a scenario is currently polling for something.
https://github.com/project-icp/bee-pollinator-app/blob/1e5820f0d064685a5175b094764cdae23a99f4b5/src/icp/js/src/compare/views.js#L205

src/icp/js/src/modeling/models.js
```diff
@@ -486,31 +486,32 @@ var ScenarioModel = Backbone.Model.extend({
...
     setResultsFromCurrentConditionsOrFetch: function() {
         var currentConditions = this.collection.models.find(function(model) {
             return model.get('is_current_conditions');
         });
-        if (currentConditions.get('results')){
-            this.set('results', currentConditions.get('results'));
-            this.set('aoi_census', currentConditions.get('aoi_census'));
-            this.set('modification_census', currentConditions.get('modification_census'));
+        if (currentConditions.get('results').models[0].get('result')) {
+            this.set('results', $.extend({}, currentConditions.get('results')));
```
arottersman: Without the copying I think any mutations on the scenario results carried over to current conditions'

src/icp/js/src/modeling/models.js
```diff
@@ -451,46 +457,79 @@ var ScenarioModel = Backbone.Model.extend({
...
-                    staleResults = inputmod_hash !== resultModel.get('inputmod_hash');
-
-                return emptyResults || staleResults;
-            });
-
-        if (needsResults && self.fetchResultsPromise === undefined) {
+            emptyResults = this.get('results').some(function(resultModel) {
+                return !resultModel.get('result');
+            }),
+            staleResults = this.get('results').some(function(resultModel) {
```
mmcfarland: I think there may still be some mutated state between scenarios.  `staleResults` for CC is inappropriately set to true if you do the following steps:
* Run CC
* Switch to New Scenario, add modification, let run
* Switch to CC

At this point it looks like `this.get('inputmod_hash')` of CC is still the correct hash, but `this.get('results')` has the `inputmod_hash` of the results returned for New Scenario

src/icp/js/src/modeling/models.js
```diff
@@ -451,46 +457,79 @@ var ScenarioModel = Backbone.Model.extend({
...
+        // Return fetchResultsPromise if it exists, else an immediately resolved one.
         return self.fetchResultsPromise || $.when();
     },
 
+    setResultsFromCurrentConditionsOrFetch: function() {
+        var currentConditions = this.collection.models.find(function(model) {
+            return model.get('is_current_conditions');
+        });
+        if (currentConditions.get('results').models[0].get('result')) {
+            this.set('results', $.extend({}, currentConditions.get('results')));
```
mmcfarland: I think the problem is here, the model in this collection is shallow copied, so the reference remains the same to the models _in_ the collection.  Doing a `currentConditions.get('results')).clone()` is still shallow.  I think we need to reconstruct the results from the original type (`ResultModel` and `ResultCollection`?) as if they were getting set from the `xhr` request that actually did get that JSON back.

arottersman: Could we also do `$.clone({}, currentConditions.get('results')) ?

arottersman: Oops, nevermind. I was testing that out on a stale bundle and thought it existed.

mmcfarland: I think this is what we need to do, the results aren't set directly so we can set them from CC `taskModel`.  If you replace the `this.set('results'...)` with:
```js
            this.set('taskModel', currentConditions.get('taskModel').clone());
            this.setResults();
```
I think it gets there, though I didn't do a full test with that result.  But it looks like it facilitates the same technique that an actual fetch would do.

PR #90, opened 2016-12-27
Add yield calculation to pollinator model

mmcfarland:
#### Overview
Computes the expected yield per cell based on algorithms provided by ICP Project members (see attached document).  Yield is a function of wild bee abundance and managed honeybee hive density, plus crop specific coefficients.

The output is also adjusted to report yield _per crop acre_ instead of per crop.  Since comparisons are made against yield types across scenarios where the specific amount of crop planted may change, we are converting the yield as calculated for a particular field to that of yield per acre of crop type on a particular field.  That way if the planted crop area is reduced, the yield per acre should be reduced in tandem, even if the change wasn't due to pollinator abundance.

Documentation: [Simple yield function for pollination DST.pdf](https://github.com/project-icp/bee-pollinator-app/files/674684/Simple.yield.function.for.pollination.DST.pdf)

Connects #https://github.com/project-icp/bee-pollinator-model/issues/2

#### Notes
Many of the crop specific data lookups are not available yet, specifically `demand` (`v` in the equations) and recommended hive density (`n` in the equations).  These are currently defaulted to arbitrary values in the appropriate order of magnitude.  The constant `k`, or `abundance_idx` is likely to stay at its current value.

It may be helpful to review the equations in person before reviewing this PR, which I'm happy to do.

#### Testing
* Reprovision your worker VM (`vagrant provision worker` to install the latest model).
* Generally, test to see if the yield implementation matches that described in the documentation.  Specifically, you might hand calculate the yield for a given cell (since it is computed per-cell), as I have done for the unit test.  If the unit test correctly checks for the right amount, it is simply aggregated across multiple cells in the raster.  
* The unit tests (on the VM, `python test/tests.py` should pass
* The `yield` endpoint should still give results to varying inputs

### DIFF
--- a/src/icp/pollinator/src/pollinator/crop_yield.py
+++ b/src/icp/pollinator/src/pollinator/crop_yield.py
@@ -1,9 +1,11 @@
 from __future__ import division
 from __future__ import print_function
 
+from scipy.ndimage.filters import generic_filter
+from collections import defaultdict
+
 from raster_ops import extract, reclassify_from_data, geometry_mask
 
-from scipy.ndimage.filters import generic_filter
 import numpy as np
 import csv
 import math
@@ -14,6 +16,7 @@ DEFAULT_DATA_PATH = os.path.join(CUR_PATH, 'data/cdl_data_grouped.csv')
 RASTER_PATH = '/opt/icp-crop-data/cdl_reclass_lzw_5070.tif'
 
 SETTINGS = {}
+ABUNDANCE_IDX = 0.1  # A constant for managing wild bee yield
 CELL_SIZE = 30
 FORAGE_DIST = 670
 AG_CLASSES = [12, 16, 17, 18, 20, 27, 33, 46, 47]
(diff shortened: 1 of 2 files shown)


src/icp/pollinator/src/pollinator/crop_yield.py
```diff
@@ -106,57 +116,95 @@ def abundance(cdl):
...
+        abundance(float): The cell value of abundance at the same position
+            as crop_id
+        managed_hives (float): Number of managed hives per acre implemented
+        config (dict): Crop specific configuration detailing `demand` the crop
+            places on bee pollination and the recommended `density` of hives
+            for that crop type
+    Returns
+        yield (float): The predicted yield for this cell position
     """
+    abundance_idx = 0.1  # A constant for managing wild bee yield
```
arottersman: Is this worth moving to the top of the file with the other constants?

mmcfarland: Yeah, I probably should.  I'll amend the commit and merge.  Thanks!

Issue #89, opened 2016-12-27
Chart crop yield output


mmcfarland:
Yield model results returns output as JSON in the format of 
```json
{
    "12": 0,
    "16": 0,
    "17": 0,
    "18": 0,
    "20": 0,
    "27": 0.8431655764579773,
    "33": 0.3809242248535156,
    "46": 0,
    "47": 0
}
```
Where the key is a CDL crop group id and the value is a measure of "yield per acre" for that crop type.  We want to display this output in a chart with the following properties:
* Vertica(?)l bar chart, one bar per crop from this scenario
* Replace crop id with crop group label (this may be done within the model, or on the client, not sure what's most beneficial)
* For crops with 0 values, don't display them
* Zip each crop for each scenario with the corresponding crop value from Current Conditions as a grouped bar.  If in that pair, one is zero and the other is non-zero, do include the crop in the chart.
* Add a class for crop id to each bar of a crop id, so we can style the bars in conjunction with the legend colors

A rough visual guide would be output like this:
![screenshot 2016-12-27 at 12 46 10 pm](https://cloud.githubusercontent.com/assets/1014341/21505177/e6681c94-cc32-11e6-92af-e1e45a5d73cb.png)

Depending on the number of classes we're reporting on, it may have to be flipped on its axis to allow longer lists of crop types.

arottersman:
Should the work on this card carry over to the compare view charts?

mmcfarland:
Good question, I don't think so.  It'll be a very similar chart, but I think we'll want to allow a crop dropdown to do one selected crop at a time (and without the CC next to the bar), similar to our compare output for mapshed.  Seems like enough to work to be its own card.

PR #93, opened 2016-12-29
Add Yield Results Chart 

arottersman:
# Overview
Up until now we've been `console.debug()`ing the yield model results and letting the old MMW tr55 runoff charts error out in the UI. At last, this PR removes the tr55 charts and adds in a yield chart!

The yield chart displays the yield per acre for each crop type.
- On scenario views,  we show the scenario's yield and the current conditions' yield for comparison. 
- On current conditions we just show the current conditions
- We only show yield for crops with either current conditions or scenario yields not equal to zero.

#### Notes
Typically grouped bar charts distinguish each group member/series by color, and then label what each is in a legend. Our bar chart, however, has each crop have its own color, and only distinguishes the series (scenario/current conditions) by shade (really fill opacity). This makes it difficult to label in the legend because there's no unifying color for all the current conditions bars, or all the scenario bars. For now, the tooltip gives the series name, but which bar is which isn't otherwise readily apparent. Should we add a pattern to one of the series so that we can add a legend, or maybe add additional labels to the bars or something?

# Demo

#### Current conditions shows single bar per crop
screen shot 2016-12-29 at 12 02 52 pm

#### Scenario shows current conditions and scenario results
screen shot 2016-12-29 at 12 03 20 pm

#### Tooltip gives value and whether bar is current Conditions/scenario
screen shot 2016-12-29 at 12 03 31 pm

# Testing Instructions
Pull this branch and `./scripts/bundle.sh --debug`
- Open dev tools - as you go through confirm there are no debug statements or errors
- Draw an AOI and get to the model view
   - most places will show "No Data"
   - `Drummond Dr, Yuba City` will have orchard/almond values
- Add land cover in current conditions and in a scenario
  - confirm the chart updates and shows different colors + appropriate labels for each crop type

Connects #89


rajadain:
> most places will show "No Data"

Is there a specific visualization for this, or just an empty tab as I see below?

image

arottersman:
Here's what it should look like if there's actually no data:
screen shot 2016-12-29 at 2 27 14 pm

rajadain:
Working now, thanks!

rajadain:
Test ran smoothly once I provisioned the worker and imported the new data. Going to look through the code now.

rajadain:
Can we make the order of the columns in the table match the order of the bars in the chart?

image

Current Conditions first, Scenario second.

src/icp/js/src/modeling/yield/views.js
```diff
@@ -0,0 +1,257 @@
...
+            this.options.scenarioResults));
+
+        return collection;
+    },
+
+    makeRowsForScenario: function(currentConditionsResults, scenarioResults) {
+        var self = this;
+        return _.map(_.keys(currentConditionsResults), function(cropType) {
+                return _.extend(self.getCropTypeValue(currentConditionsResults, scenarioResults, cropType));
+            });
```
rajadain: Some extra indentation here and the line above

src/icp/js/src/modeling/yield/views.js
```diff
@@ -0,0 +1,257 @@
...
+    initialize: function(options) {
+        this.compareMode = options.compareMode;
+    },
+
+    onAttach: function() {
+        this.addChart();
+    },
+
+    addChart: function() {
+        var formatResultsFromModel = function(results, seriesName, isCurrentConditionsData) {
```
rajadain: From a scope perspective it makes sense to sequester these functions within `addChart`, but from a performance perspective it would be better to not have to redefine them on every invocation of `addChart`.

Consider defining them as plain functions near the end of the file.

src/icp/js/src/modeling/yield/views.js
```diff
@@ -0,0 +1,257 @@
...
+            });
+        }
+
+        var chartEl = this.$el.find('.bar-chart').get(0),
+            result = this.model.get('result').quality.filter(utils.filterOutOxygenDemand),
+            aoiVolumeModel = this.options.aoiVolumeModel,
+            seriesDisplayNames = ['Suspended Solids',
+                                  'Nitrogen',
+                                  'Phosphorus'],
+            data,
```
rajadain: Consider explicitly initializing these variables to `null`

src/icp/sass/utils/_variables.scss
```diff
@@ -44,31 +47,52 @@ $sidebar-tab-subheader-height: 60px;
...
-$soil-cd: #cc4c02;
-
-//Type
-$font-family: 'Roboto', helvetica, arial, sans-serif;
-$font-h1: 1.8rem;
-$font-h2: 1.5rem;
-$font-h3: 1.0rem;
+//CDL Colors
+// TODO replace with official legend colors
+$crop-1: #5475A8;
```
rajadain: Instead of having to copy around this long list everywhere, we can reduce some redundancy by taking advantage of Sass.

Consider converting this list into a map:

```sass
$cropColors: (
  1: #5475A8,
  2: #E8D1D1,
  3: #E29E8C,
  4: #ff0000,
  5: #B50000,
  // ...
);
```

And in `_charts.scss` having

```sass
// Official CDL legend colors for classes selected for display
@for $i from 1 through 47 {
  .crop-#{$i} {
    fill: map-get($cropColors, $i)
  }
}
```

arottersman: This is so cool!

src/icp/sass/components/_charts.scss
```diff
@@ -21,53 +21,12 @@
...
-  .crop-39 { fill:  $crop-39; }
-  .crop-40 { fill:  $crop-40; }
-  .crop-41 { fill:  $crop-41; }
-  .crop-42 { fill:  $crop-42; }
-  .crop-43 { fill:  $crop-43; }
-  .crop-44 { fill:  $crop-44; }
-  .crop-45 { fill:  $crop-45; }
-  .crop-46 { fill:  $crop-46; }
-  .crop-47 { fill:  $crop-47; }
+  // Official CDL legend colors for classes selected for display
```
rajadain: This comment is repeated twice

arottersman:
Thanks for taking a look @rajadain!

Issue #95, opened 2016-12-29
Finalize CDL Colors


arottersman:
* Pull colors from original cdl and apply them to new grouped raster
* Use same colors for css classes

PR #96, opened 2016-12-30
Recolor CDL and Update CDL-Related CSS

arottersman:
# Overview
We want to apply the scheme of the original CDL colormap to the new reclassified one, but the original has some problems:
- Some crops that belong to different groups have the same color
- Some crops are an ambiguous, nasty gray that doesn't fit in with the color scheme and makes the crop look like it's "no data"

I've added a script `recolor_cdl.py` that tries to make a new colormap for each of the 47 crop groups "inspired" by the original. It 
1. removes all the gray and duplicate values from the original colormap
2. adds the not-duplicate/not-gray values using the reclass map to the new colormap by the grouped crop id
3. Some groups/keys won't have a value in the new colormap, so it uses whatever is leftover/unused from the old, not-duplicate/not-gray one to fill in the gaps

# Demo
### The raw, original CDL
screen shot 2016-12-30 at 1 17 03 pm

### The new recolored and reclassified CDL
screen shot 2016-12-30 at 1 15 58 pm

### Now with 78% more pastels
screen shot 2016-12-30 at 1 09 34 pm

# Testing Instructions
Assuming you have the two CDL tifs in your `/opt/icp-crop-data` dir, pull this branch

1. The recolor script
```
vagrant ssh worker
cd /vagrant/src/icp/pollinator/src/pollinator/reclass/
python recolor_cdl.py
```
  - When it's done you'll have `./cdl_recolor.tif`
   -   Inspect it in QGIS
   -   Temporarily replace `/opt/icp-crop-data/cdl_reclass_lzw_5070.tif` with it and confirm the model still runs

2. The app
`./scripts/bundle.sh`
- Draw an AOI 
- The landcover colors should match the new colormap's (`berries id=20, almonds=33, buckwheat=12, melons=18`) 
- The colors in the chart should match the land cover's

Connects #95


src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+        # Make sure the block size, tiling and compression
+        # is carried over to the new raster
+        src_meta.update(blockxsize=src.block_shapes[0][0],
+                        blockysize=src.block_shapes[0][1],
+                        tiled='yes', compress='lzw')
+        with rasterio.open(DST_PATH, 'w', **src_meta) as dst:
+            # Write the new colormap to the file
+            dst.write_colormap(1, create_colormap())
+            # Copy over the contents of the badly colored raster
+            for ji, window in src.block_windows(1):
```
mmcfarland: I don't think this is strictly necessary.  Since the `colormap` is a tif header attribute, we shouldn't need to move the bytes around.  We could just have the script shell out a `cp` command and work on the new file, but I don't think there's really any harm in overwriting the colormap in cdl_reclass_lzw_5070.tif directly - since we don't need it as a source for anything.  If you make that change, this script should be very fast and we'll have one less version of the raster to worry about - what do you think?

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import print_function
+from __future__ import division
+
+import numpy as np
```
mmcfarland: This import is unused.

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+from copy import copy
+
+UNGROUPED_SRC_PATH = '/opt/icp-crop-data/cdl_5070.tif'
+SRC_PATH = '/opt/icp-crop-data/cdl_reclass_lzw_5070.tif'
+DST_PATH = './cdl_recolor.tif'
+
+BUCKWHEAT = 12
+BUCKWHEAT_COLOR = (215, 158, 189)
+
+def is_gray(rgb):
```
mmcfarland: There's a few PEP8 linting warning here, mostly about number of linebreaks between `defs and some whitespace/indentation issues.

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+            for crop_id in crop_ids:
+                if crop_id in unique_old_colormap:
+                    if int(grouped_crop_id) not in colormap:
+                        # Only use first three values of the rgb -- rasterio doesn't like
+                        # the alpha value
+                        colormap[int(grouped_crop_id)] = unique_old_colormap[int(crop_id)][0:3]
+                        used_values.append(int(crop_id))
+
+        unused_unique = [value for key, value in unique_old_colormap.iteritems() \
+            if key not in used_values ]
```
mmcfarland: PEP8 rules want this to be indented to the `]` and the `\` is unnecessary 

```python
unused_unique = [value for key, value in unique_old_colormap.iteritems()
                 if key not in used_values]
```

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+        # Keep track of what colors we use, so we can use any left over colors
+        # for the unassigned ones
+        used_values = []
+        for grouped_crop_id, crop_ids in reclass_map.iteritems():
+            for crop_id in crop_ids:
+                if crop_id in unique_old_colormap:
+                    if int(grouped_crop_id) not in colormap:
+                        # Only use first three values of the rgb -- rasterio doesn't like
+                        # the alpha value
+                        colormap[int(grouped_crop_id)] = unique_old_colormap[int(crop_id)][0:3]
```
mmcfarland: I believe `crop_id` here and below are already `int` and don't need the cast

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+        colormap = {}
+        # Not all the colors in the original color map were unique to their
+        # group, so there will be some groups that don't get assigned a color.
+        # Keep track of what colors we use, so we can use any left over colors
+        # for the unassigned ones
+        used_values = []
+        for grouped_crop_id, crop_ids in reclass_map.iteritems():
+            for crop_id in crop_ids:
+                if crop_id in unique_old_colormap:
+                    if int(grouped_crop_id) not in colormap:
```
mmcfarland: Can this conditional be removed entirely if you use a `break` after setting the `colormap[grouped_crop_id]` below?  That should break out of the `for` loop and you wouldn't evaluate any additional `crop_id`s from this group.  If not, consider breaking off a part into a function both to reduce the nesting and also to introduce a named function which might enhance the clarity of the codes intentions.

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
```diff
@@ -0,0 +1,114 @@
...
+                        # Only use first three values of the rgb -- rasterio doesn't like
+                        # the alpha value
+                        colormap[int(grouped_crop_id)] = unique_old_colormap[int(crop_id)][0:3]
+                        used_values.append(int(crop_id))
+
+        unused_unique = [value for key, value in unique_old_colormap.iteritems() \
+            if key not in used_values ]
+        for grouped_crop_id in reclass_map.keys():
+            if int(grouped_crop_id) not in colormap and len(unused_unique) > 0:
```
mmcfarland: Consider casting this value once to a variable and then reusing it rather than repeating the cast.

src/icp/pollinator/src/pollinator/reclass/recolor_cdl.py
arottersman: Yeah, I'll update this - I had been doing this initially and it kept saving as a (colormapped) 0x0 file.... but now I realize I forgot to change the `'w'` flag to `'w+'`...

arottersman:
The fixups ditch the deterministic setting of the colors and instead pull them from a dictionary I tailored to look neither under, nor over-pastelized. 

Also, the recolor script now directly read/writes the reclassified file in `/opt/icp-crop-data`. It should be fast and not tamper with anything besides the colormap...

screen shot 2017-01-03 at 1 17 10 pm
screen shot 2017-01-03 at 1 16 36 pm

arottersman:
Thanks for taking a look (and sitting through the qgis mysteries)!

Issue #87, opened 2016-12-23
Share current condition modifications through all scenarios


mmcfarland:
Current Conditions is a special scenario that serves as a baseline of model results for the current land use characteristics.  It has the option exposed to adjust the land use (in practical terms, the CDL values at that location) to rectify deficiencies in the underlying model data.  Since these corrections serve as the baseline, they must be integrated with the other scenarios with speculative adjustments made to them.  Some implications:

* Land use on Current Conditions can be drawn and removed at any time
* The CC drawn shapes should also appear on other scenarios, but have a muted visualization so they aren't confused with scenario modifications.  They can't be deleted via scenario modifications, and probably shouldn't have a popup - they are the real land use.
* CC modifications should be part of the scenario data model and need to be posted to the `yield` endpoint, but should be kept separate from scenario modifications so they can be overwritten by changes from CC.
* When CC shapes are drawn or removed, they invalidate the model results of all other scenarios

Ideally, we don't fire off 5 (for example) scenario model runs for every CC shape drawn, especially as they may take some time on the CC tab to make all the corrections.  This is related to #84, where we want to manage the trigger for execution.  This might include:

* Wait until the tab for a scenario is clicked to execute
* Have non-active scenarios occasionally poll to see if their results are invalid, before they are clicked
* When the compare button is clicked, any invalid scenarios need to re-run
* Maintain a queue of requests for scenario model executions that is cleared when CC is changed, with some delay so that not all scenarios are requested at once, or may requested sequentially.

This deserves some discussion before implementation.

PR #97, opened 2016-12-30
Allow corrected land use to be applied to all scenarios

mmcfarland:
#### Overview
All modifications made or removed from the current conditions scenario are propagated across all other scenarios, including new and duplicated scenarios.  This facilitates using current conditions as a rectified base layer for constant model data. Both shared and scenario-specific modifications are submitted to the model endpoint. See detailed description in #87.

Additionally, the modification popups were altered to allow different behaviors for each type and the enhancement modifications were changed to types which have attributes more likely to affect the model output.

Connects #87 

screenshot from 2016-12-30 14 16 19

#### Known issues
There is a bug present on develop as well as this branch that cancels jobs too eagerly when switching between tabs if one is currently polling.  This can result in too many requests being generated for the number of scenarios there are.  Rather than try and fix it here, I'm going to investigate that in another issue, along with the aspects of the connected issue dealing with managing requests when modifications are shared.

#### Testing
* Update your local bundle `./scripts/bundle.sh --debug`
* Draw an area of interest and move to the modelling screen.
* To avoid the bug listed above, wait until polling completes before switching tabs, although all the intended behavior in this PR is not specifically affected - just the model results polling.
* Draw some land use changes on CC
    * Popup should be "normal", ie, let you delete
* Switch to new scenario, CC modification should be present
    * Has muted symbology
    * Popup won't let you delete, tells you where it's from
* Make new scenario, should behave like scenario above with CC changes already applied
* Duplicate a scenario, changes should be applied
* Delete a change on CC
    * Feature should be removed from all scenarios
* Scenario specific modifications should stay only available on that scenario, except when duplicated

### DIFF
--- a/src/icp/js/src/modeling/models.js
+++ b/src/icp/js/src/modeling/models.js
@@ -127,13 +127,13 @@ var ProjectModel = Backbone.Model.extend({
         this.set('is_activity', settings.get('activityMode'));
 
         this.listenTo(this.get('scenarios'), 'add', this.addIdsToScenarios, this);
-        this.get('scenarios').on('change:modification_hash', this.shareGlobalModifications);
+        this.get('scenarios').on('change:modification_hash', this.shareGlobalModifications, this);
     },
 
     shareGlobalModifications: function(changedScenario) {
         // When modifications are made to current conditions, add them
         // to all other scenarios as well
-        var scenarios = this;
+        var scenarios = this.get('scenarios');
         if (changedScenario.get('is_current_conditions')) {
             var sharedMods = changedScenario.get('modifications').toJSON();
             scenarios.each(function(scenario) {
@@ -267,6 +267,11 @@ var ProjectModel = Backbone.Model.extend({
                 });
 
             scenariosCollection.reset(scenarios);
+
+            // Apply current condition's modifications to all other scenarios
+            var currentConditions = scenariosCollection.findWhere({is_current_conditions: true});
+            this.shareGlobalModifications(currentConditions);
+
             // Set the user_id to ensure controls are properly set.
             response.user_id = user_id;
 
--- a/src/icp/js/src/modeling/models.js
+++ b/src/icp/js/src/modeling/models.js
@@ -127,6 +127,21 @@ var ProjectModel = Backbone.Model.extend({
         this.set('is_activity', settings.get('activityMode'));
 
         this.listenTo(this.get('scenarios'), 'add', this.addIdsToScenarios, this);
+        this.get('scenarios').on('change:modification_hash', this.shareGlobalModifications);
+    },
+
+    shareGlobalModifications: function(changedScenario) {
+        // When modifications are made to current conditions, add them
+        // to all other scenarios as well
+        var scenarios = this;
+        if (changedScenario.get('is_current_conditions')) {
+            var sharedMods = changedScenario.get('modifications').toJSON();
+            scenarios.each(function(scenario) {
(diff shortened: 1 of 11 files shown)


rajadain:
Since modifications drawn on Current Conditions will apply to all scenarios, consider changing this wording to say "all scenarios" instead of "this scenario"

image

rajadain:
Added modification (blue, Buckwheat) doesn't show up in the scenario's baseline in the chart or table, while basemap data (gree, Orchard and red, Almonds) do:

image

The values of basemap data is also incorrect in the new scenario.

This functionality was just added in #93 by @arottersman. She may be able to help out here, or in a future separate card.

rajadain:
If I add a new scenario after having made modifications in Current Conditions, it works correctly:

image

rajadain:
I duplicated this scenario:

image

but while it had the right shapes, it didn't have the right results:

image

rajadain:
Other tests worked well. It seems that the map / shape components are behaving as they should, but the results are not. Going to take a look at the code now.

mmcfarland:
Yeah, I suspect the results being off are a result of the polling bug that I mentioned.  I think there may be significant rework to get that all aligned and my inclination is to not muddy this PR with those changes.  If the modifications are added/removed from scenarios as they should be, I'm willing to look at the polling/results issue separately.

src/icp/js/src/modeling/models.js
```diff
@@ -127,6 +127,21 @@ var ProjectModel = Backbone.Model.extend({
...
+        this.get('scenarios').on('change:modification_hash', this.shareGlobalModifications);
+    },
+
+    shareGlobalModifications: function(changedScenario) {
+        // When modifications are made to current conditions, add them
+        // to all other scenarios as well
+        var scenarios = this;
+        if (changedScenario.get('is_current_conditions')) {
+            scenarios.each(function(scenario) {
+                var sharedMods = changedScenario.get('modifications').toJSON();
```
rajadain: Consider initializing `sharedMods` outside the `.each` loop since it'll be the same for every occurrence

rajadain:
The shapes are added / cloned / removed correctly, and the modeling runner can be fixed in a future PR.

mmcfarland:
> Will there be another card that manages saving these shared modifications to the database?

That was actually an oversight.  I'm inclined to not save them specifically and instead keep the CC scenario be the only source of truth (where they are already saved), but I'll add a commit that re-initializes the field when the project is reloaded.  Good catch!

mmcfarland:
>  consider changing this wording to say "all scenarios" instead of "this scenario"

The wording is correct when not on current conditions, but incorrect on it.  I'll add a conditional that changes it depending on the tab that is active.

mmcfarland:
Added some new commits that address the feedback, specific new behavior:
* Popups on CC for drawn modifications read differently that those for drawn modifications on non CC scenarios.
* Reloading a saved project should result in shared modifications being distributed to scenarios again.

arottersman:
Tested this pre-fixups, and it's looking good! I think I experienced the same shape/chart mismatch as @rajadain. In my case it was that I added some land cover to the current conditions after already having a few modifications. While the new mod was polling I switched to a scenario with its own existing modification. I could see the new current conditions shape, but it wasn't represented in the chart because the scenario chart didn't re-render once the current condition polling was complete. When I clicked the same scenario tab I was on again though, the chart refreshed and everything was correct.

rajadain:
Saved a project and reloaded it, the shared modifications were applied correctly. Great work! Friendly reminder to squash before merging

Issue #74, opened 2016-12-09
Add legend


ajrobbins:
Once the client has dictated the reclassification of the cropscape raster and we've integrated it in #58, add a legend. Need consultation from @jfrankl; will likely look like this example from MMW wireframes:

![screen shot 2016-12-09 at 10 28 00 am](https://cloud.githubusercontent.com/assets/12868342/21054100/3c4400ca-bdfa-11e6-919c-307d9576c98b.png)
 
(but without the layer selector)

mmcfarland:
This should include finalizing the color choices for the legend with the reclassified data.  Though subjective, I think we should use the original "primary" crop color from the grouped result since presumably some thought was put into the original color selection.

arottersman:
Will there be a way to know what crop types are visible in current viewport or will we be showing the legend with all 47 values?

arottersman:
I created https://github.com/project-icp/bee-pollinator-app/issues/95 for the actual color choosing/application

mmcfarland:
We won't be able to tell what's in the viewport, I think we should just do all 47 for the purpose of this card, but it's likely we'll do something like show the top 10 most common/important and then do a "show more" for the remainder, but we'll need to talk to @jfrankl for guidance.  This card can just be roughing it in.

PR #99, opened 2017-01-03
Add Crop Layer Legend

rajadain:
## Overview

Adds a legend for the crop layer, visible whenever the crop layer is visible. This is an initial stab. We need to replace "Crop n" labels with proper names, and better manage the number of items.

### Demo

image

### Notes

This uses placeholder colors. Actual colors are expected in #95 / #96. When the variables are updated, this should update automatically.

## Testing Instructions

Pull down this branch. Run the bundle script. Open the app. Enable the crop layer. You should see a legend right above the crop layer button. It should go away when the crop layer is hidden.

Connects #74

### DIFF
--- a/src/icp/js/src/core/overlayControl.js
+++ b/src/icp/js/src/core/overlayControl.js
@@ -4,7 +4,8 @@ var L = require('leaflet'),
     $ = require('jquery'),
     _ = require('underscore'),
     Marionette = require('../../shim/backbone.marionette'),
-    overlayControlTmpl = require('./templates/overlayControl.html');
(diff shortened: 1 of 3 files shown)


arottersman:
https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/js/src/core/cropTypes.json might be helpful when you're ready to add in the crop names

rajadain:
@arottersman neat, I'll incorporate those

rajadain:
Added names in f307e75

image

rajadain:
We'll figure out the column spacing in the second pass. Thanks for taking a look!

mmcfarland:
New commit looks good.

Issue #94, opened 2016-12-29
Add yield chart to compare view


arottersman:
See https://github.com/project-icp/bee-pollinator-app/issues/89#issuecomment-269662588

PR #100, opened 2017-01-04
Add yield results chart to Compare view

jtarricone:
This adds stacked bar charts to the Compare view, for the baseline and for each scenario.

For example, if the basic modelling view looks like this:
download 10

The Compare view now looks like this:
download 11

To test:
* clone this branch and open the app
* draw some AoI that has nonzero crop data somewhere within it and create at least one scenario with some set of modifications
* when the model has completed, open the "Compare" view and check that the stacked bar charts have been drawn and look accurate

Connects #94

### DIFF
--- a/src/icp/js/src/compare/controllers.js
+++ b/src/icp/js/src/compare/controllers.js
@@ -5,7 +5,8 @@ var _ = require('lodash'),
     router = require('../router').router,
     views = require('./views'),
     modelingModels = require('../modeling/models.js'),
-    modelingControls = require('../modeling/controls');
+    modelingControls = require('../modeling/controls'),
+    cropTypes = require('../core/cropTypes.json');
 
 var CompareController = {
     compare: function(projectId) {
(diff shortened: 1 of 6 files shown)


mmcfarland:
Looks like there was some confusion on the issue - the original issue indicates that the crop should be selectable to show one crop at a time across scenarios, not a stacked bar of all the crops.  For one, it's hard to compare across scenarios to get the value for a particular type if it's stuck in the middle of the stack, but probably more importantly we're not really summing up these relative yield/acre/crop units to a single value - which was the case for MMW runoff.

jtarricone:
That would be a single-item bar in each scenario, right?

mmcfarland:
Yep, one bar/one crop, with the dropdown that currently reads "Crop yield" being a list of the union of any crops in any scenarios (meaning, if berries is just in one scenario, it can still be selected but other scenarios will hav a zero height bar for it).

jtarricone:
I pushed an update that brings the implementation closer to what I think the target is. I added some commentary to explain the reasoning where it seemed useful.

arottersman:
Going to pull this down now -- will post updated screenshots as I test

src/icp/js/src/modeling/models.js
```diff
@@ -81,19 +81,31 @@ var ResultCollection = Backbone.Collection.extend({
         return this.findWhere({name: name});
     },
 
+    getResultByAttribute(attribute, value) {
+        var d = {};
+        d[attribute] = value;
+        return this.findWhere(d);
```
jtarricone: I think JS chokes on any indirection in object keys, and `attribute` is a variable in this case.

src/icp/js/src/core/chart.js
```diff
@@ -32,6 +32,121 @@ function getNumBars(data) {
...
+// Therefore, we manually remove the tooltip when elements under chartEl are
+// destroyed.
+function removeTooltipOnDestroy(chartEl, tooltip) {
+    $(chartEl).children().bind('destroyed', function() {
+        $('#' + tooltip.id()).remove();
+    });
+}
+
+/*
+    Renders a stacked vertical bar chart for multiple series of data
```
jtarricone: It is, I copied the entire section from the MMW implementation. The description is still accurate, but the Compare view only charts a "series of one"

src/icp/js/src/compare/controllers.js
```diff
@@ -75,6 +75,60 @@ function setupProjectCopy(aoi_census) {
...
+            var results = r.get('result');
+            var clones = _.map(results, function (value, key) {
+                var clone = r.clone();
+                clone.set('result', { key: key, value: value });
+                clone.set('displayName', cropTypes[key]);
+                return clone;
+            });
+            return clones;
+        });
+        return _.head(transformedModels);
```
jtarricone: I'll add a blurb.

src/icp/js/src/modeling/yield/views.js
```diff
@@ -185,17 +183,31 @@ var CompareChartView = Marionette.ItemView.extend({
...
+
         $(chartEl).empty();
         if (result) {
-            data = null;
-            chartOptions = null;
+            data = getData(result),
+            chartOptions = {
+                yAxisLabel: 'Yield Per Acre',
+                barClasses: data.class,
+                yAxisUnits: 'bees',
```
jtarricone: Not exactly, I didn't know what the real units were so I needed a placeholder.

src/icp/js/src/modeling/yield/views.js
```diff
@@ -185,17 +183,31 @@ var CompareChartView = Marionette.ItemView.extend({
...
+                margin: {top: 20, right: 0, bottom: 40, left: 60},
+                showLegend: false,
+                disableToggle: true,
+                abbreviateTicks: true
```
jtarricone: (Possibly) good catch, I was wondering what was causing that. It only seems to appear for empty series, there could be a default unit D3 sets somewhere.

src/icp/js/src/modeling/yield/views.js
arottersman: I think the option key is `yAxisUnit` though

jtarricone:
I pushed an additional commit that updates the chart y-axis domain and permits styling, as well as fixups to address the other comments.

src/icp/js/src/core/chart.js
```diff
@@ -32,6 +32,143 @@ function getNumBars(data) {
...
+            $(bar).attr('style', '');
+        });
+    }
+
+    function updateChart() {
+        if($svg.is(':visible')) {
+            setChartWidth();
+            chart
+                .staggerLabels($svg.width() < widthCutoff)
+                .update(); // Throws error if updating a hidden svg.
```
arottersman: I think we also need to call `addBarClasses()` here after the update

src/icp/js/src/modeling/models.js
arottersman: Oh, yeah, no ES6 :(

Issue #98, opened 2016-12-30
Refactor how scenarios poll for model results


mmcfarland:
Currently, there is some special logic new scenarios that share `inputmod_hashes` with current conditions and also some special logic around `shared_modifications` that essentially allow a scenario to be edited when it is not active.  This has accumulated into some incorrect behavior around when results are polled and especially cancelled.  Refactor the polling so that it more comprehensively allows for this behavior.

Specifically, results could be cached by `inputmod_hash` which would return either the results as is or a promise to an in-progress poll.  This would also remove extra model runs when a modification is added and then removed, since the previous state's hash results would still be available.

mmcfarland:
Spent some time tracing down the "Draw modification before original results are returned" bug with not much success - looks pretty complicated.  I'm going to create an issue for that and the similar compare view issue, can this get reviewed without that?

PR #103, opened 2017-01-06
Consistent polling behavior

mmcfarland:
#### Overview
With the changes introduced for editing current conditions (CC), sharing modifications and the workflow that assumes the user will spend some time on CC, there were some inconsistent behavior introduced for when a scenario would poll for results, when that polling would be cancelled and when that polling would be diverted and just use an existing scenario's result. This PR adds additional guards to make sure that scenarios poll for results as expected

Connects #98 

#### Testing
Generally, adding modifications or changing inputs on any scenario should ultimately update that scenario, without interference from work on other scenarios. Some specifics to check:
* Multiple scenarios can be edited sequentially without waiting for polling to complete on any given one
* Switching tabs while scenarios are polling should not alter their in-progress polling
* When a scenario is equivalent to CC, it should quickly just display those pre-computed results
* When modifying CC and there are several scenarios, those scenarios don't poll until they are activated.
* Reloaded a saved project should not reinstate any polling for new results (assuming you let the results complete before reloading)
* Doing sequential modification on a scenario, without letting the polling complete, should cancel the previous polling and start a new one.  The "polling" spinner should also remain visible after the second polling sequence has begun, as appropriate.

### DIFF
--- a/src/icp/js/src/core/models.js
+++ b/src/icp/js/src/core/models.js
@@ -85,7 +85,7 @@ var MapModel = Backbone.Model.extend({
 
 var TaskModel = Backbone.Model.extend({
     defaults: {
-        pollInterval: 500,
+        pollInterval: 1000,
         /* The timeout is set to 45 seconds here, while in the
            src/icp/apps/modeling/tasks.py file it is set to 42
            seconds.  That was done because the countdown starts in the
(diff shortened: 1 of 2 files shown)


mmcfarland:
Spent some time tracing down the "Draw modification before original results are returned" bug with not much success - looks pretty complicated.  I'm going to create an issue for that and the similar compare view issue, can this get reviewed without that?

arottersman:
Sounds good - I'll take a look now!

Issue #107, opened 2017-01-12
Add "All" option to Crop Type dropdown in Compare View


mmcfarland:
Since most users will have < 4 crops in a field at any given time, we can add an All option to the crop picker to allow multiple crop bars to be shown, per scenario.

PR #113, opened 2017-01-20
Allow Compare All

mmcfarland:
#### Overview
Adds an "All Crops" option to the compare chart, to view all crops with values in the AoI.  The chart is generally optimized for showing up to 4 crop types, which is the upper end of what actual users might have.  Things tend to get squished above that value, but the client thinks the vast majority will have only two crop types at a time.  Also, tries to simplify the compare chart by reusing the existing chart rendering code and not modifying the format of the `ResultsModel`.

Connects #107 

screenshot from 2017-01-20 10 53 11
screenshot from 2017-01-20 10 53 27

#### Testing
* Rebundle the front end assets `./scripts/bundle.sh --debug`
* Generate model results for a field that contains multiple crops
* In compare view, verify you can view all (up to 4) crops at once and still have the ability to choose single crop types.

### DIFF
--- a/src/icp/js/src/compare/controllers.js
+++ b/src/icp/js/src/compare/controllers.js
@@ -4,8 +4,7 @@ var _ = require('lodash'),
     App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
-    modelingModels = require('../modeling/models.js'),
-    cropTypes = require('../core/cropTypes.json');
+    modelingModels = require('../modeling/models.js');
 
 var CompareController = {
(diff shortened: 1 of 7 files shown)


mmcfarland:
Unclear re: tablets.  I don't think so, primarily, and the 4 crop was meant to be an upper limit.  We agreed to add this feature with the caveat that the chart may get squished, in which case they user will just use the single crop options.  That was sort of the tradeoff of implementing this, so I'm just going to let it be for now.

Issue #115, opened 2017-01-24
Error when loading results in compare view


lewfish:
If you trigger an initial model run and then switch to the compare view before the results have returned, the following error message appears in the console, the spinner keeps going, and the new results never appear. This is true even after the results being returned by the backend. This also happens when triggering a second model run in the modeling view, and then switching to the compare view, although the error in the console doesn't appear.

```
views.js:689 Uncaught TypeError: Cannot read property 'show' of undefined
    at child.showDetailsRegion (http://localhost:8000/static/js/main.js:8551:32)
    at Array.<anonymous> (http://localhost:8000/static/js/main.js:8533:18)
    at fire (http://localhost:8000/static/js/vendor.js:42654:30)
    at Object.fireWith [as resolveWith] (http://localhost:8000/static/js/vendor.js:42766:7)
    at Array.<anonymous> (http://localhost:8000/static/js/vendor.js:42894:16)
    at fire (http://localhost:8000/static/js/vendor.js:42654:30)
    at Object.fireWith [as resolveWith] (http://localhost:8000/static/js/vendor.js:42766:7)
    at Array.deferred.(anonymous function) (http://localhost:8000/static/js/vendor.js:42856:34)
    at fire (http://localhost:8000/static/js/vendor.js:42654:30)
    at Object.fireWith [as resolveWith] (http://localhost:8000/static/js/vendor.js:42766:7)
```

lewfish:
This doesn't happen in MMW.

rajadain:
Here's what I've discovered so far. In the Compare view, even after polling finishes successfully, the actual value of the `results[n].polling` isn't updated:

![image](https://cloud.githubusercontent.com/assets/1430060/22480604/f8cc48ee-e7bf-11e6-964c-dce3dd635330.png)

This is confusing, because `setPolling(false)` executes correctly. This does not happen if we stay on the modeling view. It is possible this is a side effect of #103.

rajadain:
Okay, confirmed that the spinners that don't stop also don't stop in MMW. This seems to be caused by the copy of the project that is made before entering Compare view.

When we started wanting to run models from the Compare view in https://github.com/WikiWatershed/model-my-watershed/pull/673, we wanted to do it in a way that did not affect the state of the Modeling view. Thus, we make a sandboxed copy of the project before handing it to the Compare view.

[...]

Issue #105, opened 2017-01-12
Allow Current Conditions to be modified before the initial model results return


mmcfarland:
There is a bug involving a cancelled job preventing the model results region from ever showing if it is cancelled before results are shown for the first time.

lewfish:
This also happens in MMW in the modeling view.

PR #116, opened 2017-01-24
Use model changes to update results view

lewfish:
Previously, making a modification to a scenario while initially polling for results would result in an error. This PR fixes this bug by switching from using a promise to using model events to update the UI. See commit message for more details.

#### Testing
* Check that the bug cannot be replicated on this branch.
* Cause a timeout error to occur and check that timeout error message is shown. You can do this by changing the constant at  https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/js/src/core/models.js#L95 
* Cause a network error to occur by turning off the network in developer tools and check that the generic error message is shown.
* With results showing, trigger a model run and check that the current results are shown while polling, and that the new results appear. 
* Repeat above while logged in.
* Repeat above in the compare view, although there is a pre-existing bug that may come up documented in https://github.com/project-icp/bee-pollinator-app/issues/115

Connects #105


lewfish:
> if you switch to compare view while the app is polling, there's a spinner that doesn't go away and I don't think the results update

See https://github.com/project-icp/bee-pollinator-app/issues/115

lewfish:
> If you are already offline and try to kick off a job, you will be stuck in "Calculating results" forever
> (if you become offline in the middle of polling, there's an error as expected)

Good catch. I added a commit to fix this.

lewfish:
> Also, (in a separate issue) we should consider adding a "retry" button on error. Currently there's no way to re-kick off a job without updating the modifications

Nice idea. Feel free to make an issue for this on Bees and MMW.

Issue #123, opened 2017-01-25
Increase click target on crop layer opacity slider


arottersman:
You can only change the slider value by dragging the thumb circle or clicking the very thin track:
![screen shot 2017-01-25 at 2 15 29 pm](https://cloud.githubusercontent.com/assets/7633670/22305232/c0d3d90e-e308-11e6-84f5-45b172129f5d.png)

We should increase the hit target.

Possible solution:

Add the following [here](https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/sass/components/_slider.scss#L15)
```
    border-top: 11px solid transparent;
    border-bottom: 11px solid transparent;
```

arottersman:
Shares point with https://github.com/azavea/pwd-parcel-viewer/issues/468

PR #125, opened 2017-01-27
Increase opacity slider click area + IE11 fix

arottersman:
### Overview
Connects #123 

This PR adds a transparent border to the opacity slider to make it easier to click. It also fixes the styling on IE11 which previously looked like this (it's missing entirely):

screen shot 2017-01-26 at 5 43 39 pm

### Demo

opacityslider

### Testing instructions
Pull and bundle
- Test that the opacity slider looks okay and is easy to click (ie has a large hit target) on Chrome, Firefox, IE11 and Safari


arottersman:
Thanks for reviewing!

Issue #124, opened 2017-01-26
Allow decimal input for Managed Hives


mmcfarland:
Manged hives can be put in as a fractional number (2.3, etc) and the UI, endpoint and model should all accept it.  Currently it either doesn't trigger a model run, or is converted into an int.

PR #126, opened 2017-01-27
Decimal hive inputs

mmcfarland:
### Overview
Allows input for "number of hives" of single digit decimals, from 0 to 10.

Connects #124 

#### @ 1.2 h/a
screenshot from 2017-01-27 14 55 01

#### @ 1.3 h/a
screenshot from 2017-01-27 14 55 13

screenshot from 2017-01-27 14 58 43

### Testing
* Rebundle the client JS
* Attempt run the model with number of hives expressed as a decimal
* Verify that `/yield` requests include the decimal values
* Verify that the model output for two scenarios which share the same whole number, but different decimals, differ slightly.

### DIFF
--- a/src/icp/js/src/modeling/templates/controls/userInput.html
+++ b/src/icp/js/src/modeling/templates/controls/userInput.html
@@ -6,6 +6,7 @@
         id="{{ userInputName }}"
         value="{{ value }}"
         min="0"
+        step="0.1"
     >
     <div class="info-region"></div>
 </div>
--- a/src/icp/js/src/modeling/controls.js
+++ b/src/icp/js/src/modeling/controls.js
@@ -237,7 +237,7 @@ var NumberOfHivesView = ControlView.extend({
     },
 
     onInputChange: function() {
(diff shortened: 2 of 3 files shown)


src/icp/js/src/modeling/templates/controls/userInput.html
```diff
@@ -6,6 +6,8 @@
         id="{{ userInputName }}"
         value="{{ value }}"
         min="0"
+        max="10"
```
arottersman: Did you want to prevent input above 10? Currently it just maxes out the increment button.

screen shot 2017-01-27 at 4 11 20 pm

mmcfarland:
I'm ok with leaving them the ability to type in whatever they want. There's no real problem with high values, it's just more likely that they will stop improving that model results.

mmcfarland:
Actually, I think for that same reason, I'll take out the max from the increment as well.  That's part of the information in the output, if one scenario has 3 h/a and the other 20 h/a, but they both have the same yield output, you begin to understand that you don't need all those extra hives.

Issue #121, opened 2017-01-24
Add recommended managed hive density to crop data


mmcfarland:
A new attribute will need to be appended to the cdl data for managed hive density.  Depending on the form it takes, we may need to script this in or manually update the source CSV.  Once all crops (including synthesized) have density values, update the model to use the value per crop instead of the static placeholder value.

mmcfarland:
Shares a point with #118

Issue #118, opened 2017-01-24
Keep certain crops explicit when grouping/reclassifying CDL


mmcfarland:
We have a special case for Almonds in the reclass scripts, which prevent it from being absorbed into its group.  We'll need to ensure that the following types are handled in the same way, ensuring they have their own entry in the CDL data after aggregation/reclassification.  For context, these are the types which will be allowed for land cover modifications.

Almond
Apple
Blueberry
Cherry
Raspberry
Pumpkin
~~Watermelon~~ (Already a single type)
~~Grassland/pasture~~ (Not a crop type, so does not need separation for density or demand values)

PR #129, opened 2017-01-30
Add more crop groups

mmcfarland:
### Overview
Updates the source CDL csv to allow for more distinct groups, resulting in a larger number of crops in the CDL. This is done to accommodate specific values for hive density for crops that would otherwise be grouped together (cherries vs orchard).  Note we only have density numbers for a few crop types. Some new crop types are added, which are derived from a set of current crop types.  They are the original + a cover crop, these have different HN/HF values from their base.

Connects #118 
Connects #121 

#### Source for types and density
`Spreadsheet can be found on the fileshare`

screenshot from 2017-01-30 20 26 12
screenshot from 2017-01-30 20 25 49

### Testing
* Test creating new grouped csv `python  group_cdl_data.py`
  * New crop types + cover crop types should be added
  * Test that the output is the version that is included in this PR
  * Check that the mapping json also updated to reflect the changes described
* Reclassify the raster based on the new grouped csv
  * Check that the distinct types have been preserved in the new raster
    * For example, Blueberry (`57`) is it's own type, and is predominant around Hammonton, NJ
* Recolor the reclassified raster
  * Since these colors have lost some meaning, it is sufficient to check that the script still runs correctly


mmcfarland:
Thanks!  Unfortunately, on today's call I got some clarifying information that will result to changes in this approach. I'm going to merge this and follow up with a PR that makes some targeted changes.

Issue #127, opened 2017-01-30
Add support for large raster on Worker instances


hectcastro:
The `Worker` instances require a ~2GB raster file to function properly when deployed to AWS. Determine the best method of getting this file into the AMIs.

mmcfarland:
Raster is available at ` s3://icp-cdl-data/cdl_reclass_lzw_5070.tif`, in the stg account.

mmcfarland:
Additionally, the yield model has a default expectation on the path

 https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/pollinator/src/pollinator/crop_yield.py#L16

which can be overridden by the calling client code

https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/pollinator/src/pollinator/crop_yield.py#L194

We can extract that out to an ENV variable to accommodate this task.

mmcfarland:
Closed via #136

PR #130, opened 2017-01-30
WIP: Add support for building Packer AMIs

hectcastro:
Make changes necessary to support building AMIs for the following machine types:

- Monitoring
- Application
- Worker


deployment/packer/template.js
```diff
@@ -38,14 +38,6 @@
             "instance_type": "t2.large",
             "ssh_username": "ubuntu",
             "ami_name": "icp-worker-{{timestamp}}-{{user `version`}}",
-            "ami_block_device_mappings": [
```
hectcastro: This may come back after we come up with a solution for https://github.com/project-icp/bee-pollinator-app/issues/127.

Issue #119, opened 2017-01-24
Report on combination of crop + synthesized crop


mmcfarland:
Some reportable crop types will have a connected sibling "crop type" that is synthesized with "Cover Crop":

Almond with cover crop
Apple with cover crop
Blueberry with cover crop
Cherry with cover crop
Raspberry with cover crop
Watermelon with cover crop
Pumpkin with cover crop

These will be options for land cover correction (and have different HN/HF/Density values), however, when reporting yield, these should be aggregated some that Almond + Almond w/ CC are reported as just Almond.  This could be done on the client or model side, so long as the average is maintained - but may be cleaner to produce output that is just by ag type.

mmcfarland:
Crops do report on:
#### crop
Almond
Apple
Blueberry
Cherry
Raspberry
Watermelon
Pumpkin
Cucurbits
Melons
Berries
Orchard
Strawberries

PR #139, opened 2017-02-01
Update Land Cover Mods and Report Cover Crops With Main Crop

arottersman:
# Overview

- Replaces placeholder land cover mods with a more finalized list (bye, bye buckwheat 🌱 You will be missed )
- Modifies the model to aggregate crops over the updated land cover mods and each of their cover crop types (model only sends back results under the main crop's id)

Connects #119

#### Notes
Updated `cropTypes.json` to fix the chart titles. The legend consumes it, too. We may want to filter it or something to get rid of the cover crops which won't appear in the raster:
screen shot 2017-02-01 at 4 28 33 pm

Left updating the colors for a time when things are more settled
screen shot 2017-02-01 at 4 28 37 pm

# Demo

screen shot 2017-02-01 at 5 03 00 pm

Just copy and pasted whatever colors for testing purposes:
screen shot 2017-02-01 at 5 14 22 pm

# Testing instructions

* pull branch 
* reinstall python model `vagrant ssh && sudo python /vagrant/src/icp/pollinator/setup.py install`
* rebundle
* Draw an area of interest in the app and confirm all crops and their crop types get grouped together in the charts


src/icp/pollinator/src/pollinator/crop_yield.py
```diff
@@ -19,7 +19,8 @@
 ABUNDANCE_IDX = 0.1  # A constant for managing wild bee yield
 CELL_SIZE = 30
 FORAGE_DIST = 670
-AG_CLASSES = [12, 16, 17, 18, 20, 27, 33, 46, 47]
+# key = main crop id to aggregate under: value = cover crop id
+AG_CLASSES = { 17: 18, 21: 22, 29: 30, 32: 33, 39: 40, 55: 56, 57: 58 }
```
mmcfarland: This is a good strategy, but I don't think will work with the full list of reportable crop types.  I've updated the issue with the list of all crop types that we're reporting yield on, some of which don't have cover crop associations.  You may need to alter the data structure to accommodate those.

Since you're altering these values here, I think it makes sense to also add include that full reportable list now.  To complicate matters, the IDs will need to reference those from #140, so you may want to see that get merged and rebase on it.

mmcfarland:
Good call about removing the cover crops from the legend, I think you should add a commit.  It's crude, but I think we can do a simple filter on the text `with cover crop` to determine inclusion.

src/icp/pollinator/src/pollinator/crop_yield.py
arottersman: I probably shouldn't have assumed they'd all have cover crops...will rebase on #140 and make some adjustments!

arottersman:
- I got tired of copy+pasting from the grouped csv into `cropTypes.json` on every update, so a made  a script that pulls out the mapping. Can remove that commit if there won't be much use for it.
    - I ended up manually removing the `cover crops` and `enhancement` types from the output `json` We don't ever need them there because they won't show up in the chart anyway.
- Rebased on develop and updated the aggregate crops/crops to report on list to match those in the fileshare spreadsheet + new ids
- Added finalized enhancement list to front end
- Added a commit to show a different message for an enhancement's popup (before it assumed everything was landcover)

###  New enhancement list, orchard, almonds, and pumpkins on the map
screen shot 2017-02-02 at 12 50 28 pm

### Updated enhancement popup
screen shot 2017-02-02 at 1 23 07 pm

src/icp/pollinator/src/pollinator/crop_yield.py
```diff
@@ -173,9 +178,12 @@ def aggregate_crops(yield_field, cdl_field, crops=AG_CLASSES):
 
     # Average the yield for each each crop type cell, by crop
     for crop in crops:
-        # Create a mask for values that are not this crop type and include
+        # Create a mask for values that are not this crop type, (or, if it
+        # has a paired crop, its pair), and include
         # the mask which is already applied to non-field areas of AoI
-        cdl_mask = np.ma.masked_where(cdl_field != crop, cdl_field).mask
+        paired_crop = paired_crops[crop] if crop in paired_crops else crop
+        cdl_mask = np.ma.masked_where((cdl_field != crop) & (cdl_field != paired_crop),
```
mmcfarland: The double mask on the same crop _probably_ doesn't incur any penalty, but it sort of sticks out.

```python
crop_mask = cdl_field != crop
if crop in paired_crops:
    crop_mask = crop_mask & (cdl_field != paired_crop)

cdl_mask = np.ma.masked_where(crop_mask, cdl_field).mask
```

This is slightly more explicit and I think equivalent, but I can't tell which I prefer - doesn't look like a significant improvement.  What do you think?

mmcfarland: This new functionality is important enough to warrant a new unit test.  Consider adding a test that specifically confirms that a paired type is aggregated correctly.  Should be easy to contrive since you made `paired_crops` an optional argument.  Should be pretty similar to `ModelTests.test_aggregation`.

arottersman: I think I prefer your suggestion -- it requires less thinking than trying to figure out why a `paired_crop` would be the crop itself.

arottersman: Will do!

src/icp/pollinator/tests/tests.py
```diff
@@ -108,6 +108,24 @@ def test_aggregation(self):
...
+        # Expect the values to be the mean yield value per crop + its paired crop
+        # if it has one.
+        # Crops that were masked out should be 0
+        expected = {
+            '100': 0,
+            '200': 1.6 / 5 * 100,
+            '300': .9 / 3 * 100,
+            '400': 0
+        }
+
```
mmcfarland: Looks like the test is missing an assertion at the end.

src/icp/pollinator/tests/tests.py
```diff
@@ -108,6 +108,24 @@ def test_aggregation(self):
...
+        crops = [100, 200, 300, 400]
+        paired_crops = { 200: 400, }
+        yield_by_crop = aggregate_crops(self.yield_field, self.cdl, crops, paired_crops)
+
+        # Expect the values to be the mean yield value per crop + its paired crop
+        # if it has one.
+        # Crops that were masked out should be 0
+        expected = {
+            '100': 0,
+            '200': 1.6 / 5 * 100,
```
mmcfarland: Since `400` is the paired crop, I don't think this tests the pairing, since "types 100 and 400 are completely outside of unmasked field area".  I think if the pair is `{ 200: 300 }`, then you can test the output to make sure it doesn't report _any_ `300` crop since they should have been included in the `200`.  And the `200` should report `(0.2 * 2 + 0.3 * 3) / 5 * 100` or `26`.  I guess technically, it will still report `300` because that is in the `crops` list.  I'd suggest removing it with a comment about how it is paired.

src/icp/pollinator/tests/tests.py
```diff
@@ -108,6 +108,24 @@ def test_aggregation(self):
         self.assertEqual(yield_by_crop, expected,
                          "Crop yield amounts were not aggregated correctly")
 
+
```
mmcfarland: A little bit of whitespace lint has been added with this function.

arottersman:
Sorry, that was pretty absent-minded of me. fixup c1bec5c makes the test actually test stuff and removes the lint.

src/icp/pollinator/tests/tests.py
```diff
@@ -108,23 +108,25 @@ def test_aggregation(self):
...
 
-
     def test_aggregation_with_paired_crops(self):
         # Crop types 100 and 400 are completely outside of unmasked field area
-        crops = [100, 200, 300, 400]
-        paired_crops = { 200: 400, }
+        # 300 is in the field area, but we pair it with 200 instead of
+        # aggregating over it
+        crops = [100, 200, 400]
+        paired_crops = { 200: 300, }
```
mmcfarland: Unfortunately, this extra white space is also lint - the rest of the file is tidied up in #148.

arottersman:
Thanks for taking a look! Squashed and removed the excess whitepsace.

Issue #141, opened 2017-02-02
Compare View Issues


rajadain:
When opening the Compare view immediately after the Modeling view, without giving the Current Conditions scenario run to complete, initially we see the spinners. Subsequently, after the spinning stops, we see this:

![image](https://cloud.githubusercontent.com/assets/1430060/22531632/2fa241ca-e8b0-11e6-9bdf-b2a59dd9e540.png)

This is likely because we [only update the charts once the spinning stops](https://github.com/project-icp/bee-pollinator-app/blob/c4a5cad8f0ab80a4d5fa8e156ca8ba3a26be042e/src/icp/js/src/compare/views.js#L217-L220), but if they are not initialized with the right variables for charting it doesn't render correctly.

When the view is reinitialized, for example by going back to Modeling and coming back to Compare, the chart renders correctly.

This can be seen in action here:

![2017-02-01 18 46 36](https://cloud.githubusercontent.com/assets/1430060/22531688/879d064e-e8b0-11e6-9cfc-c52575c86dca.gif)

Another issue is that we seem to not update scenarios until they have been made active. Thus, after having made changes to Current Conditions, if the user goes directly to Compare view, other scenarios will be stale. But if they go to each scenario first and then go to the Compare view, the results are accurate.

![2017-02-01 19 03 58](https://cloud.githubusercontent.com/assets/1430060/22531799/5658db5c-e8b1-11e6-8a88-60dec225276d.gif)

A third issue is that if any changes in the Modeling view lead to additional chartable crops being added, or some removed, and the user goes to the Compare view before results come back, then Compare view isn't updated correctly. It keeps the list of chartable crops [during initialization](https://github.com/project-icp/bee-pollinator-app/blob/c4a5cad8f0ab80a4d5fa8e156ca8ba3a26be042e/src/icp/js/src/compare/views.js#L181-L189), and doesn't update them.

rajadain:
One of the main causes for these failures is the behavior introduced in https://github.com/project-icp/bee-pollinator-app/pull/103, namely that of stale scenarios not updating themselves until their tab is made active. This leads to the Compare View having stale results, as described here, until that scenario's tab is viewed first in Model view.

I can envision two approaches to fix this:

 1. Going to Compare view forces recalculation of every scenario that is out of date.
    + ➕  Requires no additional user input
    + ➕  With individual spinners for each scenario it is plainly communicated that they are updating.
[...]

Issue #45, opened 2016-11-14
Clean up Python Code


rajadain:
Currently there are both linting and test failures:

```
> kj test
+ vagrant ssh app -c 'flake8 /opt/app/apps --exclude migrations || echo flake8 check failed'
/opt/app/apps/user/views.py:6:1: F401 'User' imported but unused
/opt/app/apps/user/views.py:7:1: F401 'reverse' imported but unused
flake8 check failed
Connection to 127.0.0.1 closed.
+ vagrant ssh app -c 'cd /opt/app && envdir /etc/icp.d/env ./manage.py test --noinput'
Creating test database for alias 'default'...
..EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE.......F.
```

Clean it up. Also see #28 for a JavaScript version of this card.

Issue #28, opened 2016-11-09
Clean up codebase/get rid of linting errors


arottersman:
```
> jshint $(find js -name *.js -not -path '*/shim/*')

js/src/analyze/views.js: line 5, col 5, 'L' is defined but never used.
js/src/analyze/views.js: line 7, col 5, 'Backbone' is defined but never used.
js/src/analyze/views.js: line 10, col 5, 'models' is defined but never used.
js/src/analyze/views.js: line 12, col 5, 'modalModels' is defined but never used.
js/src/analyze/views.js: line 13, col 5, 'modalViews' is defined but never used.
js/src/analyze/views.js: line 15, col 5, 'coreViews' is defined but never used.
js/src/analyze/views.js: line 17, col 5, 'chart' is defined but never used.
js/src/analyze/views.js: line 18, col 5, 'utils' is defined but never used.
js/src/analyze/views.js: line 20, col 5, 'analyzeResultsTmpl' is defined but never used.
js/src/analyze/views.js: line 24, col 5, 'barChartTmpl' is defined but never used.
js/src/analyze/views.js: line 52, col 13, 'projectUrl' is defined but never used.

js/src/app.js: line 3, col 5, '$' is defined but never used.
js/src/app.js: line 7, col 5, 'settings' is defined but never used.

js/src/core/utils.js: line 4, col 5, 'lodash' is defined but never used.

js/src/core/views.js: line 10, col 5, 'coreUtils' is defined but never used.
js/src/core/views.js: line 191, col 13, 'self' is defined but never used.

js/src/draw/tests.js: line 135, col 10, 'assertTextEqual' is defined but never used.

js/src/draw/views.js: line 24, col 10, 'actOnUI' is defined but never used.
js/src/draw/views.js: line 35, col 10, 'actOnLayer' is defined but never used.
js/src/draw/views.js: line 66, col 10, 'validateClickedPointWithinDRB' is defined but never used.
js/src/draw/views.js: line 225, col 10, 'makePointGeoJson' is defined but never used.

js/src/modeling/controllers.js: line 5, col 5, 'Backbone' is defined but never used.
js/src/modeling/controllers.js: line 188, col 9, 'modelPackageDisplayName' is defined but never used.

js/src/modeling/views.js: line 7, col 5, 'csrf' is defined but never used.

24 errors
```

rajadain:
Created #45 for cleaning up Python code.

mmcfarland:
Shares a point with #45

PR #142, opened 2017-02-02
Use same project for Compare view as Modeling

rajadain:
## Overview

Previously, we would create a sandboxed copy of the main project for the Compare view. This was done in order to preserve the state of the main project from any changes made in the Compare view, which could trigger model runs and modify input values.

This led to bugs such as spinners continuing to spin if that's how Compare view was initialized, even after any polling finishes. The original project's results would be updated, but not of the sandboxed copy, causing the user to see stale results.

Since we don't expect the Pollinator App to be able to trigger model runs from the Compare view, it is safe to remove the sandboxed copy of the project and just use the original one. This simplifies the code, making it easier to reason about, as well as removing any bugs like the one mentioned above.

This functionality was originally added in Model My Watershed in https://github.com/WikiWatershed/model-my-watershed/pull/673

Connects #115 

### Notes

I found a number of quirks with the Compare view, which I've recorded in #141. Please read that issue before reporting the same quirks here.

## Testing Instructions

 * Check out this branch and update bundle scripts
 * Create a new project. Wait for Current Conditions and New Scenario to process in the Modeling view.
 * Make a change in New Scenario and switch to Compare view before processing finishes. Ensure that the Compare view's chart updates with the new values, and the spinner stops accurately.
 * Ensure you are logged out. Draw a shape and proceed to the Compare view. Log in from the Compare view. Ensure that your project is saved correctly.

### DIFF
--- a/src/icp/js/src/compare/controllers.js
+++ b/src/icp/js/src/compare/controllers.js
@@ -1,38 +1,24 @@
 "use strict";
 
-var _ = require('lodash'),
-    App = require('../app'),
+var App = require('../app'),
     router = require('../router').router,
     views = require('./views'),
     modelingModels = require('../modeling/models.js');
 
 var CompareController = {
     compare: function(projectId) {
-        var first = null,
-            aoi_census = null;
-
         if (App.currentProject) {
-            first = App.currentProject.get('scenarios').first();
-            if (first) {
-                aoi_census = first.get('aoi_census');
-            }
-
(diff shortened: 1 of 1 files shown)


rajadain:
@azavea-bot rebuild

rajadain:
The checks fail because of linting errors, for which we have cards #28 and #45. Merging this now. Thanks for taking a look!

Issue #138, opened 2017-02-01
Impose maximum field size


mmcfarland:
The model was designed for use on field sizes up to 400 acres, we should enforce a limit near that value for drawn areas.

arottersman:
Updating the max to 410 acres is just a one-line change -- should its work include porting over the MMW fancy modals?
![screen shot 2017-02-02 at 3 08 16 pm](https://cloud.githubusercontent.com/assets/7633670/22566356/903af654-e959-11e6-9bbd-1e778eed5ec1.png)

mmcfarland:
Bummer that we didn't capture that before we cut over.  I think it should be relatively simple to copy over the nicer modal, let's do it unless it begins to appear challenging.

arottersman:
For max field size should it be exactly `400`? And should it be inclusive or exclusive?

mmcfarland:
Might as well keep it at 400 inclusive, if we need to tweak it we can, we can also check the impact on execution time if they find they need it bigger.

PR #147, opened 2017-02-02
Max draw area size update  + fancy modals

arottersman:
## Overview

Model only intended for fields of up to 400 acres, so the max size has been updated.

We were using `window.alert()` to show the error message, so I've ported over the MMW modals.

Connects #138 

### Too big
screen shot 2017-02-02 at 5 15 38 pm

### Duplicate scenario name
screen shot 2017-02-02 at 4 28 51 pm

## Testing instructions 
* Pull this branch and bundle up
* Draw an AOI > ~5/8ths of a square mile
   - should see error
* Draw good, little sized AOI, should succeed
* Rename scenarios to duplicate names/make stuff fail -- confirm there are no more `window.alert`

### DIFF
--- a/src/icp/js/src/draw/views.js
+++ b/src/icp/js/src/draw/views.js
@@ -18,7 +18,8 @@ var $ = require('jquery'),
     drawTmpl = require('./templates/draw.html'),
     resetDrawTmpl = require('./templates/reset.html'),
     windowTmpl = require('./templates/window.html'),
-    settings = require('../core/settings');
+    settings = require('../core/settings'),
+    modalViews = require('../core/modals/views');
 
 var MAX_DRAW_ACRES = 400; // About 5/8's of a square mile
 var codeToLayer = {}; // code to layer mapping
@@ -50,14 +51,12 @@ function validateShape(polygon) {
         var errorMsg = 'This watershed shape is invalid because it intersects ' +
                        'itself. Try drawing the shape again without crossing ' +
                        'over its own border.';
-        window.alert(errorMsg);
         d.reject(errorMsg);
     } else if (area > MAX_DRAW_ACRES) {
         var message = 'Sorry, your Area of Interest is too large.\n\n' +
                       Math.floor(area).toLocaleString() + ' acres were selected, ' +
                       'but the maximum supported size is currently ' +
                       MAX_DRAW_ACRES.toLocaleString() + ' acres.';
-        window.alert(message);
         d.reject(message);
     } else {
         d.resolve(polygon);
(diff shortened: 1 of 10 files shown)


src/icp/js/src/modeling/models.js
```diff
@@ -775,13 +777,18 @@ var ScenariosCollection = Backbone.Collection.extend({
         });
 
         if (match) {
-            window.alert("There is another scenario with the same name. " +
-                    "Please choose a unique name for this scenario.");
-
-            console.log('This name is already in use.');
+                    console.log('This name is already in use.');
```
rajadain: Extra couple of indents in this entire block.

src/icp/js/src/modeling/views.js
```diff
@@ -179,9 +179,16 @@ var ProjectMenuView = Marionette.ItemView.extend({
             var xhr = self.model.destroy({wait: true});
             if (xhr) {
                 xhr.done(navigateAfterDelete)
-                    .fail(function() {
-                        window.alert('Could not delete this project.');
+                .fail(function() {
```
rajadain: Consider identing this block one level in, to maintain visual parity with `.done` in the previous line

src/icp/js/src/draw/views.js
```diff
@@ -161,8 +171,11 @@ var DrawWindow = Marionette.LayoutView.extend({
                     isDrawing: false,
                     isDrawn: true
                 });
-            }).fail(function() {
+            }).fail(function(message) {
                 revertLayer();
+                if (message) {
+                    displayAlert(message, modalModels.AlertTypes.error);
```
rajadain: Since `displayAlert` is only ever used once, consider simply inlining it here rather than declaring and calling a function.

arottersman:
Thanks, Terence! I think simplifying the alert modal rendering is a great idea, and have added fixup commit to do so. 

I removed the `window.alert` from the tests -- I haven't checked if my change passes though...

rajadain:
This looks much nicer! I'm not sure what's the difference between an Error and Warning here, since in both cases we reprimand the user for the wrong behavior (selecting too big an area, naming a scenario incorrectly), show them the right thing to do, and undo their action (area selection, scenario renaming).

Let's use Error for application errors ("An error occurred", "Could not duplicate this project"), and Warn for user errors ("AoI too big", "Scenario name already exists").

src/icp/js/src/draw/views.js
```diff
@@ -175,8 +160,11 @@ var DrawWindow = Marionette.LayoutView.extend({
                     isDrawing: false,
                     isDrawn: true
                 });
-            }).fail(function() {
+            }).fail(function(message) {
                 revertLayer();
+                if (message) {
+                    modalViews.showError(message);
```
rajadain: Consider switching this to `showWarning` since it is a user error.

arottersman:
Thanks for reviewing! Squashed fixups and will merge.

Here's the AOI message as a warning:

screen shot 2017-02-03 at 11 15 56 am

Issue #158, opened 2017-02-09
Add map controls


mmcfarland:
The basemap selector and overlay control are not visible in production.  The locate me control is in production, but not in development.

PR #159, opened 2017-02-09
Have consistent map controls in both development and production

mmcfarland:
Makes sure that the basemap selector, overlay control and geolocation buttons are all consistently placed in both environments.  Previously, development had the layers but not the location and production had the location but not the layers.

#### Before
screenshot from 2017-02-09 09 27 19

#### After
screenshot from 2017-02-09 09 28 07

Connects #158 

#### Testing
* Verify that development control show correctly
* Check production settings are updated to show the layer controls


mmcfarland:
Note that the location control won't function in Chrome on development as Google requires secure origins to allow location access.

Issue #156, opened 2017-02-08
Update colors


mmcfarland:
* Make sure that colors in the legend, chart and vector drawing are lined up again.
* Give a consistent border color to the land use/enhancements for in-scenario versions (dark)

I believe that the CDL ids are stable again, so we should probably re-associate the meaningful colors for common types:
* Reportable crops
* Land cover values
* Enhancement values
* Developed (low, med & high can use shades of the same, preferably gray or understated).

PR #161, opened 2017-02-13
Unify color definitions for CDL types

mmcfarland:
### Overview
Move to a single definition of crop colors based on CDL id.  Define this once with a class and then use everywhere where a crop color is needed, to allow easy changes to colors in the future and ensure consistency.  The colors remain defined in `cdl_colormap.py`, which is used for raster recoloring, but are output as hex for use in the scss.  Special attention was paid to crops which are reported on or used to correct a project, but they are identified so updates or tweaks can happen pretty easily.

Connects #156 

### Demo
screenshot from 2017-02-13 09 50 03
screenshot from 2017-02-13 09 49 50
screenshot from 2017-02-13 09 49 23

### Testing
* Rebundle `./scripts/bundle.sh --debug`
* Verify that all instances of a crop being visualized uses the same color from the same single source.
  * Yield chart
  * Thumbnail selector
  * Legend
  * Drawn Vector

I did spend some time picking sensible default colors for the "important" CDL types, but I'm open for these to be tweaked when a final style pass is given to the app.


src/icp/js/src/draw/utils.js
```diff
@@ -5,8 +5,7 @@ var $ = require('jquery'),
     _ = require('lodash');
 
 var polygonDefaults = {
-        fill: false,
-        color: '#f9cf2f'
+        color: '#212121'
```
arottersman: Was this updated to read better on lighter colored fields? It now is hard to see on darker colored ones:

screen shot 2017-02-13 at 2 43 56 pm

vs 

screen shot 2017-02-13 at 2 44 44 pm

src/icp/pollinator/src/pollinator/reclass/cdl_colormap.py
```diff
@@ -9,53 +11,59 @@
...
+    55: (48, 63, 159),     # Blueberry CC
+    56: (183, 28, 28),     # Cherry CC
+    57: (239, 83, 80),     # Raspberry CC
+    58: (240, 98, 146),    # Watermelon CC
+    59: (245, 124, 0),     # Pumpkin CC
+    60: (225, 190, 231),   # Wildflower early
+    61: (171, 71, 188),    # Wildflower late
+    62: (230, 238, 156),   # Woody early
+    63: (192, 202, 51),    # Woody late
+    64: (129, 212, 250),   # Mix early
```
arottersman: Mix early and mix late are pretty similar (at least on my monitor)
screen shot 2017-02-13 at 2 38 30 pm

src/icp/pollinator/src/pollinator/reclass/cdl_colormap.py
```diff
@@ -9,53 +11,59 @@
...
+    65: (79, 195, 247),    # Mix late
```
arottersman: Mix early and mix late are pretty similar (at least on my monitor)
screen shot 2017-02-13 at 2 38 30 pm

src/icp/js/src/draw/utils.js
mmcfarland: Nope,  good catch - that was an unintended consequence.  I was focusing on the modifications and changed the field polygon collaterally.  Will restore.

mmcfarland:
Updated to restore yellow selection color for AoI and used the same difference for ealy/late mix that I used for the other enhancement types.

screenshot from 2017-02-13 16 11 27

PR #165, opened 2017-02-16
Always Compare Fresh Scenarios

rajadain:
## Overview

Addresses various issues of the compare view resulting from loading it before all scenarios are done refreshing. Also does some minor refactoring of the fetch code.

See original issue for a detailed description (with animated GIFs) of the problems.

Connects #141

### Notes

There is an interesting (good) side effect of the refactoring: when deleting modifications from a New Scenario so that it is identical to Current Conditions, it no longer polls for results, instead just copies the results from Current Conditions instantly.

When polling does finish in Compare view, all dropdowns will be reset to "All Crops" since a crop the dropdown may have been previously set to may no longer be available.

## Testing Instructions

 * Check out the branch and bundle your scripts `kj bundle --debug`
 * Open the app and ensure you are not logged in

 1. Test that going straight to Compare view from Draw / Modeling without waiting for Current Conditions to finish works correctly
    * This should trigger New Scenario to fetch results as well
    * When Current Conditions returns, its graph and dropdown should be correctly populated
    * When New Scenario returns, its graph and dropdown should be correctly populated
 2. Test that making a change to a scenario (by adding or deleting a modification or input) and moving to the Compare view before results come back works correctly.
    * No additional fetches should be fired
    * When results come back, the proper scenario should be updated
 3. Add a modification to a New Scenario. Wait for it to finish polling. Then go to Current Conditions and add a modification. Go to Compare view before polling finishes.
    * Ensure both Current Conditions and New Scenario are polling, since changes to Current Conditions affects New Scenario as well, and it can't simply copy over Current Conditions since it has modifications of its own
    * Ensure the values are updated correctly
 4. Completely cover New Scenario with a new crop type that isn't listed in the results. Go to Compare view before polling finishes.
    * Ensure when polling finishes, all charts (New Scenario and any siblings) show all crops, including the new one that come back from this fetch.
    * Ensure that all dropdowns list all available crops

 * Test this any other way you can think of

Also tagging @mmcfarland who may want to take a cursory glance.

### DIFF
--- a/src/icp/js/src/compare/utils.js
+++ b/src/icp/js/src/compare/utils.js
@@ -1,0 +1,18 @@
+"use strict";
+
+var _ = require('lodash');
+
+function getChartableCrops(scenarios) {
+    return scenarios.reduce(function(cropTypes, scenario) {
+        var nonZeroCrops = scenario.get('results').map(function(result) {
+            return Object.keys(_.omit(result.get('result'), function(value) {
+                return value === 0;
+            }));
+        });
+        return _.uniq(cropTypes.concat(_.flatten(nonZeroCrops)));
+    }, []);
+}
+
+module.exports = {
+    getChartableCrops: getChartableCrops,
+};
--- a/src/icp/js/src/compare/views.js
+++ b/src/icp/js/src/compare/views.js
@@ -45,6 +45,8 @@ var CompareWindow = Marionette.LayoutView.extend({
         // so the offset of the container needs to be
         // recomputed.
         $(window).bind('resize.app', _.debounce(_.bind(this.updateContainerPos, this)));
+
+        this.model.fetchResultsIfNeeded();
     },
 
     onDestroy: function() {
--- a/src/icp/js/src/compare/views.js
+++ b/src/icp/js/src/compare/views.js
@@ -9,6 +9,7 @@ var _ = require('lodash'),
     coreViews = require('../core/views'),
     modelingViews = require('../modeling/views'),
     modConfigUtils = require('../modeling/modificationConfigUtils'),
+    utils = require('./utils'),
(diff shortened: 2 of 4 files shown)


rajadain:
Interesting to see the order of commits listed in GitHub is different (and incorrect) from what I have:

```
* 0917aea - Update Compare View on polling finish (24 hours ago) <Terence Tuhinanshu>
* 361f977 - Fetch results when loading Compare view (24 hours ago) <Terence Tuhinanshu>
* cdc1ed7 - Simplify fetch logic (24 hours ago) <Terence Tuhinanshu>
* 1ddff11 - Refactor fetching logic while keeping behavior (24 hours ago) <Terence Tuhinanshu>
```

Probably because GitHub uses `AuthorDate`, not `CommitDate` when listing them:

```
> git show --pretty=fuller 0917aea
commit 0917aeaf33e850fff1319d2c771df98857a7761d
Author:     Terence Tuhinanshu <ttuhinanshu@azavea.com>
AuthorDate: Wed Feb 15 17:42:49 2017 -0500
Commit:     Terence Tuhinanshu <ttuhinanshu@azavea.com>
CommitDate: Wed Feb 15 18:14:45 2017 -0500

    Update Compare View on polling finish
```

arottersman:
Taking a look at this now!

rajadain:
Can't see the map in your screenshot, just confirming: the area with cherries on Current Conditions wasn't occluded by Apples in New Scenario, was it? That could explain it.

arottersman:
That may have been the case, but I hit the wrong the wrong back button and lost the project 😬 . Now we'll never know

rajadain:
Got it. We can make a new issue if it crops up again. Thanks for taking a look and thinking through the refactoring!

Issue #163, opened 2017-02-15
Unaccounted for CDL value


mmcfarland:
Rollbar captured an error that implies a CDL value in the raster not accounted for in `cdl_data`, resulting in a `KeyError`.  Maybe a `NODATA` value? The following json was the supplied input to the exception:

https://rollbar.com/ProjectICP/Bee-Pollinator/items/3/

```
[{"inputs": [{"name": "numberofhives", "area": "0", "effectiveUnits": null, "value": 0, "summary": "", "shape": null, "effectiveShape": null, "units": "acres", "effectiveArea": null, "type": ""}], "modification_hash": "6c09ce1a35a198c89a54192fec45a465", "aoi_census": null, "modification_censuses": null, "modification_pieces": [{"category": "landcover", "shape": {"geometry": {"type": "Polygon", "coordinates": [[[-86.28597736358643, 42.36636043798232], [-86.28659963607787, 42.36626531376108], [-86.28670692443848, 42.364743306631716], [-86.28741502761841, 42.3647750154897], [-86.2877368927002, 42.3643786536149], [-86.28780126571655, 42.36271390643378], [-86.28677129745483, 42.362539501795474], [-86.28679275512695, 42.364220108164794], [-86.28563404083252, 42.36423596272781], [-86.28559112548827, 42.36645556205954], [-86.28597736358643, 42.36636043798232]]]}, "type": "Feature", "properties": {}}, "value": 51, "area": 9.737636078497584}], "area_of_interest": {"type": "MultiPolygon", "coordinates": [[[[-86.28788709640502, 42.36689947251543], [-86.28372430801392, 42.36695496110185], [-86.2855589389801, 42.362238256279966], [-86.28784418106079, 42.36228582145739], [-86.28788709640502, 42.36689947251543]]]]}, "inputmod_hash": "22f18ea54d980ee1b6ce6739762de08b6c09ce1a35a198c89a54192fec45a465"}]
```

arottersman:
Here's a project with the drawn shape: https://pollinator-modeling-app.azavea.com/project/17/scenario/35

PR #167, opened 2017-02-17
Account for NODATA values in yield calculation

mmcfarland:
## Overview

When `NODATA (0)` values are present in the CDL the model now assumes there is no yield for that cell. This scenario would be encountered when drawing an area on an a lake, or near a lake when the entire "beeshed" was taken into account.

The previous behavior was to generate an exception on trying to look up crop configuration based on the `NODATA` value.  A unit test has been added to verify this fix.

Connects #163 

## Demo
screenshot from 2017-02-17 17 11 10

## Testing
* Install the updated model `vagrant ssh worker -c "cd /opt/app/pollinator && sudo python setup.py install"`
* Restart celery if needed, or `./scripts/debugcelery.sh`
* Search for Wildwood, Michigan and find the farm in the screenshot, just to the east of the geocoded point.  In fact, you can draw anywhere in that region that within ~2km of Lake Michigan (which contains the `NODATA` cells).  That farm initiated the first observation for the error.
* Run a model in that area, get expected results (non-error, there don't happen to be registered crops at that location)
* Ensure new tests passes

### DIFF
--- a/src/icp/pollinator/tests/tests.py
+++ b/src/icp/pollinator/tests/tests.py
@@ -145,6 +145,22 @@ class ModelTests(unittest.TestCase):
         self.assertEqual(cell_yield, 0.886,
                          "Yield not calculated correctly for psuedo-crop 1")
 
+    def test_nodata_yield(self):
+        """
+        Test that NODATA values are handled properly.
+        """
+        config = {
+            1: {'demand': 0.95, 'density': 2.5}
+        }
+
+        nodata = 0
+        abundance_idx = 0.4
+        managed_hives = 1
+        cell_yield = yield_calc(nodata, abundance_idx, managed_hives, config)
+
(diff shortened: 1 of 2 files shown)


src/icp/pollinator/tests/tests.py
```diff
@@ -145,6 +145,22 @@ def test_yield(self):
...
+            1: {'demand': 0.95, 'density': 2.5}
+        }
+
+        nodata = 0
+        abundance_idx = 0.4
+        managed_hives = 1
+        cell_yield = yield_calc(nodata, abundance_idx, managed_hives, config)
+
+        self.assertEqual(cell_yield, 0,
+                         "NOATA should return 0 value for yield")
```
arottersman: Typo

mmcfarland:
Thanks, fixed the label typo.

Issue #166, opened 2017-02-17
Error when setting Managed Hives Per Acre to null


arottersman:
Currently you can delete the contents of the managed hives per acre input:
![screen shot 2017-02-17 at 10 41 22 am](https://cloud.githubusercontent.com/assets/7633670/23071825/a557d5a2-f4fd-11e6-93e6-65882c3565cf.png)

Then on blur this causes the modeling job to fail, and the compare view graph to be blank with the hives/acre as `NaN`:
![screen shot 2017-02-17 at 10 38 17 am](https://cloud.githubusercontent.com/assets/7633670/23071944/0addfbc2-f4fe-11e6-96dc-ea364bb72a22.png)

rajadain:
What would be _more_ correct? Setting the value to 0, or to what it previously was before clearing it, if a user erases it? I'm leaning more towards the latter /cc @ajrobbins

ajrobbins:
If the user was trying to erase it, I think it should reset to 0.

rajadain:
I was thinking that if the user wants to set it to 0, they can manually do so. If, on the other hand, they accidentally erase it, then the new value (0 in case we do that) will be saved to the project immediately, and with no undo option, the old value will be lost.

Setting it to the old value ensures there are no accidents, and if they want to make it 0 they can do so manually.

PR #168, opened 2017-02-20
Set Number of Hives to 0 upon erasure

rajadain:
## Overview

If the user erases the value of Number of Hives, we take it to mean "Reset" and set the value to 0.

The first commit was an attempt to discard the empty value, and revert to the previous value, assuming the erasure was accidental. I'll squash it before merging.

Connects #166 

## Testing Instructions

 * Check out this branch and bundle scripts `./scripts/bundle.sh --debug`
 * Get to the Modeling view, and erase the value of Number of Hives. Ensure that it updates to "0" when you leave the text box.
 * Set the value to 2. Wait for the model run to finish. Then erase the value. Ensure that since the previous value was different, it triggers a model run.
 * Erase the value again and leave the text box again. Ensure that since the previous value was 0, it does not trigger a model run.
 * Ensure that setting values normally via keyboard and via the up/down arrows works.
 * Ensure the Compare view works correctly.

### DIFF
--- a/src/icp/js/src/modeling/controls.js
+++ b/src/icp/js/src/modeling/controls.js
@@ -235,11 +235,14 @@ var NumberOfHivesView = ControlView.extend({
     },
 
     onInputChange: function() {
-        var value = parseFloat(this.ui.input.val());
+        var value = parseFloat(this.ui.input.val()) || 0;
+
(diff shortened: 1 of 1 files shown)


arottersman:
Pulling this down now!

Issue #170, opened 2017-02-20
Use client supplied GA keys


mmcfarland:
The app is currently associated with incorrect Google Analytics, update the production settings with a new key.

@ajrobbins this is mostly for you to reach out and have them set up an account and also give us access to it.  We need the `UA-*` key and it should be (for now) associated with the `https://pollinator-modeling-app.azavea.com` domain, though the production one to come.

ajrobbins:
Oooh I get a card! Fun! Will reach out.

mmcfarland:
We've still go to update the value, but we've already got a spot for it.

ajrobbins:
key is `UA-92733102-1`

rajadain:
@ajrobbins In case you'd like to do this yourself, it's a simple change to this line: https://github.com/project-icp/bee-pollinator-app/blob/d9ab5a631b9a87e5db75bbf969bf969995a9e8b9/src/icp/icp/settings/production.py#L111

ajrobbins:
The client confirmed that GA is running and showing usage statistics.

PR #174, opened 2017-03-21
Add ICP google analytics account

arottersman:
Connects #170 
@ajrobbins ... is there a way to test this?


ajrobbins:
This diff looks good to me. I don't think we can actually test until it's in use on staging, at which time I'll ask the Bees folks if their GA appears to be showing accurate analytics.

arottersman:
Thanks. Going to merge it then.

Issue #188, opened 2017-04-20
Remove "early" options from Enhancements dropdown and rename


ajrobbins:
Related to https://github.com/project-icp/bee-pollinator-app/issues/187, remove the "early" options from the Enhancements dropdown:
![screen shot 2017-04-20 at 11 37 29 am](https://cloud.githubusercontent.com/assets/12868342/25239344/c13d5d96-25bd-11e7-9cb1-9e8c0104d6dc.png)

Rename the "late" options to the following (and keep colors):
- Wildflower
- Woody
- Wildflower-Woody Mix

We should probably rearrange the dropdown to have three horizontal options, instead of vertical.

arottersman:
Shares a point with #183

Issue #183, opened 2017-04-20
Make entire "Crop Layer" button clickable


ajrobbins:
Currently, you have to click on the eye icon within the Crop Layer button to turn on the CDL overlay. We should make the entire button clickable.

![screen shot 2017-04-20 at 11 19 58 am](https://cloud.githubusercontent.com/assets/12868342/25238454/7d881de0-25bb-11e7-97ba-fae3c41f7328.png)

arottersman:
Shares a point #188

PR #199, opened 2017-04-25
Visual Fixups

rajadain:
## Overview

Fixes minor visual issues. See commit messages for details.

Connects #183 
Connects #188 
Connects #200 

### Demo

image

## Testing Instructions

 * Check out this branch and `./scripts/bundle.sh --debug`
 * Open the app. Ensure you can enable the crop layer in a much easier way from before. Ensure that other related controls, such as the opacity slider and legend toggle, still work as before.
 * Go to the modeling stage. Check the Enhancements under New Scenario. Ensure that Early variants have been removed, and Late variants have been renamed and rearranged.


rajadain:
> Noticed #200 while testing.

Good catch, will do.

rajadain:
Just pushed up a commit for it.

Demo:

rajadain:
Thanks for taking a look!

Issue #179, opened 2017-04-20
Compare graph overlaps text below


ajrobbins:
On compare page:
![screen shot 2017-04-20 at 11 11 00 am](https://cloud.githubusercontent.com/assets/12868342/25238000/2c513eee-25ba-11e7-8eac-68c3e1107b61.png)

arottersman:
What should happen if the view is too short for the modifications to show at all?

![screen shot 2017-04-25 at 5 06 25 pm](https://cloud.githubusercontent.com/assets/7633670/25407785/a83c6f24-29d9-11e7-9cec-3c2812b39132.png)

I was thinking the whole panel could scroll:

![ibwplwclud](https://cloud.githubusercontent.com/assets/7633670/25407847/f0703956-29d9-11e7-911f-4446924a2925.gif)

ajrobbins:
I like the scrolling! @jfrankl ?

arottersman:
@jfrankl — I have a branch with this change if you want to take this behavior out for a test run.

PR #204, opened 2017-05-02
Fix compare view overlap on short viewports

arottersman:
### Overview

The compare view charts overlapped with the modeling info below them on short viewports:
image

If we move the modifications table down, it will be entirely off the screen on shorter viewports, so this PR just sets the entire compare column to scroll.

It also makes some styling updates per @jfrankl's suggestion: the spacing around the modification was unnecessarily large. I've also centered the "Navigate Scenarios" text on the back arrow.

Connects #179 

### Demo
hv2qavh0sr

### Testing instructions
- Pull and `./scripts/bundle.sh` (make sure updated css files get written)
- Draw an AoI over somewhere with data (ie the triangle near `yuba city`)
- Go to "Compare"
- Confirm view is useable at a variety of window sizes

### DIFF
--- a/src/icp/sass/pages/_compare.scss
+++ b/src/icp/sass/pages/_compare.scss
@@ -1,10 +1,15 @@
+$compare-footer-height: 50px;
+$compare-map-height: 25vh;
+// Calculate the height between the header and footer, and below the map
+$compare-chart-table-height: calc(100vh - #{$height-lg} - #{$compare-footer-height} - #{$compare-map-height});
+
 .compare-scenarios-container {
   overflow: hidden;
   position: absolute;
   top: $height-lg;
   left: 0;
   right: 0;
-  bottom: 50px;
+  bottom: $compare-footer-height;
   z-index: 100;
   background-color: $ui-light;
   transition: 0.3s ease left, 0.3s ease right;
@@ -15,7 +20,7 @@
     .scenario-overlay {
       position: absolute;
       z-index: 100;
-      height: 22%;
+      height: $compare-map-height;
       width: 100%;
       background-color: black;
       opacity: 0.25;
@@ -68,7 +73,7 @@
           }
 
           .map-region {
-            height: 22%;
+            height: $compare-map-height;
             width: 100%;
 
             .map-container {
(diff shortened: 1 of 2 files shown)


rajadain:
Taking a quick look in IE and FF as well

rajadain:
FF works well. On IE11 in Windows, while there is no more overlap, there are no scrollbars either, making the content inaccessible:

image

rajadain:
Scrollbars _are_ visible on FF on Windows

image

Perhaps a little unseemly, but functional.

rajadain:
Also on Chrome on Windows

image

arottersman:
Fixup:
- I was calculating the chart+table height with a nested `calc()` (`$compare-height` mapped to another `calc`). IE11 couldn't figure it out, so I removed the intermediary variable
- Added `overflow-x: hidden` to the chart-table to remove the unseemly double scrollbar 
screen shot 2017-05-03 at 9 38 27 am


P.S. Where did you find olives??

rajadain:
> Where did you find the olives??

Just hadn't reprovisioned my VM in a while. They're gone now.

IE11 now has correct sidebars, and they work well. Unseemly as the horizontal sidebars may have been, they did serve a purpose. When the screen is too narrow, they help make hidden content accessible, like in this case:

image

I think rather than setting `overflow-x: hidden`, we should figure out a better width for those columns. However, since this is likely to involve some experimentation, I'd suggest making another card for it so it can be pointed and accomplished on its own.

src/icp/sass/pages/_compare.scss
```diff
@@ -105,15 +110,20 @@
             }
           }
 
+          .chart-table-regions {
+              height: $compare-chart-table-height;
+              overflow-y: auto;
+              overflow-x: hidden;
```
rajadain: Remove this line.

mmcfarland:
> I think rather than setting overflow-x: hidden, we should figure out a better width for those columns. 

A deep reach back to https://github.com/project-icp/bee-pollinator-app/pull/113 reveals some earlier discussion where we confirmed with the client that the compare chart is optimized for the likely max of 4 crops, and beyond that the individual crop dropdown will be used.  We shouldn't need to optimize the width for compare any further.

arottersman:
Thanks for your thorough testing, @rajadain! Squashed fixups. Going to merge. Jeff mentioned off hand that it's a little ridiculous to have the barchart bars so wide. If we ever address that, it may help with blueberries getting cut off in your screenshot above.

Issue #205, opened 2017-05-02
Perform model output comparison and validation


mmcfarland:
We supplied model input parameters to the ICP team who then produced a farm abundance output from their internal model tool.  We'll compare their output to that of our tool and note whatever calibrations or adjustments are necessary.

mmcfarland:
Swapped for https://github.com/azavea/usace-levee-safety-api/issues/71

PR #206, opened 2017-05-02
Add debug output for rasters

mmcfarland:
## Overview
To help validate the model and share results, a method which can be
enabled manually has been added, which saves intermediate rasters to
disk.

Connects #205 

## Notes
Using these saved rasters, I compared our output to that supplied by the ICP team.  At `2000m` buffer that we've been using, there is some noticeable misalignment of values, although they are distributed very similarly.  Increasing the buffer to `5000m` produced nearly identical results.  The comparison can be seen here: 
[Pollinator DST Model comparison.pdf](https://github.com/project-icp/bee-pollinator-app/files/971504/Pollinator.DST.Model.comparison.pdf)

Additionally, even though the farm abundance layer is created with differences at the two buffer values, there is a very marginal net change to the yield calculation output.

#### 5km buffer
Crop | Yield
----|----
Orchard | 65.408
Almonds |	64.404

#### 2.5km buffer
Crop | Yield
----|----
Orchard	| 65.365
Almonds	| 64.44

We've provided this information and the tifs to the ICP team for them to evaluate the appropriate buffer value to use.

## Testing
* Change `DEBUG=True`
* Update the model installation `sudo python setup.py install` or `vagrant provision worker`
* Verify that the rasters are saved to disk
* Verify they are not saved when `DEBUG=False`

### DIFF
--- a/src/icp/pollinator/src/pollinator/crop_yield.py
+++ b/src/icp/pollinator/src/pollinator/crop_yield.py
@@ -4,12 +4,16 @@ from __future__ import print_function
 from scipy.ndimage.filters import generic_filter
 from collections import defaultdict
 
-from raster_ops import extract, reclassify_from_data, geometry_mask
+from raster_ops import extract, reclassify_from_data, geometry_mask, \
+    write_tif
 
 import numpy as np
 import csv
 import math
 import os
+
+# Enableing DEBUG will write all intermediate rasters to disk
+DEBUG = False
 
 CUR_PATH = os.path.dirname(__file__)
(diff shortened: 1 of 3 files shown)


mmcfarland:
Got it, will update the tests to account for these new signatures.  

Re: color, I didn't think it too important to add.  The CDL is thematic and has a scheme, but the other files would just be a ramp based on min/max, which qgis will apply be default (grayscale).  I applied a uniform min/max ramp over the files in my report so the files with different actual min/max values could be visually compared easily.

mmcfarland:
Tests updated and have passed on CI.

Issue #210, opened 2017-05-04
Add text when no data available


ajrobbins:
When no data is available, add text that says "Correct land cover to see yield results."

mmcfarland:
Shares a point with #202

Issue #202, opened 2017-05-01
Replace favicon


ajrobbins:
[Bee-silhouette6.ico.zip](https://github.com/project-icp/bee-pollinator-app/files/967929/Bee-silhouette6.ico.zip)

PR #216, opened 2017-05-05
Favicon and No Data

mmcfarland:
## Overview

When there are no yield values to report on, don't include the bar
chart which comes with its own "no data" text.  Instead, render our own
custom messages via app templates. The logic includes the ability to
include extra information only when yield is visible, including the
table and an explanation.

Also updates the favicon to ICP provided image.

Connects #210 
Connects #202 

## Demo
**Favicon**
screenshot from 2017-05-05 15 45 08

**Scenario no-data**
screenshot from 2017-05-05 15 21 05

**Scenario yes-data**
screenshot from 2017-05-05 15 20 51

**Compare yes and no-data**
screenshot from 2017-05-05 15 20 39

## Testing
* Bundle, `scripty manage collectstatic` and refresh
* In the desert or an urban area where there are unlikely to be any crops, add a field
* Verify that the appropariate "No Data" messages are shown in Scenario tabs and Compare
* Verify that the chart, table and "relative yield" description are not show
* Add a crop to a scenario (or redo the area in a cropped area) and verify that the no data message are not shown and the data is.
* Verify the bee silhouette is the new favicon


rajadain:
Taking a look at this now.

PR #238, opened 2017-06-28
Style updates

### LINKED ISSUE #237: UI issues
Feedback from client:

- [x] Remove or hide up/down arrows next to managed honey bee input box

- [x] Add y axis labels to graphs that appear in the compare window (e.g. label y axis  with "% of maximum yield")  

- [x] Make the 'calculating' spinning symbol a bright color so it is more visible.  

- [x] Change the color of the yellow text in the right frame (e.g. "crop yield" and baseline  and scenario yield numbers in the table). Color should change from yellow to a more  visible color given the white background.

jfrankl:
Fixes #237.

### Add y axis to graphs in compare view

image

### Remove up and down arrows on number input

image

### Change yellow text color to brown so it's easier to read

image

### Change color for calculating spinner

screen shot 2017-06-28 at 1 03 00 pm

### DIFF
--- a/src/icp/js/src/modeling/yield/views.js
+++ b/src/icp/js/src/modeling/yield/views.js
@@ -221,7 +221,6 @@ var CompareChartView = Marionette.ItemView.extend({
                     yAxisUnit: '% of Maximum Yield',
                     barClasses: _.pluck(_.flatten(_.pluck(data, 'values')), 'class'),
                     maxBarWidth: 100,
-                    margin: {top: 20, right: 0, bottom: 40, left: 40},
                     showLegend: false,
                     disableToggle: true,
                     yAxisDomain: [0, 100],
--- a/src/icp/sass/base/_typography.scss
+++ b/src/icp/sass/base/_typography.scss
@@ -59,3 +59,7 @@ label {
     font-size: 0.5rem;
   }
 }
+
+a {
+  color: $brand-secondary;
+}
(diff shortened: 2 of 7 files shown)


rajadain:
Taking a look now.

rajadain:
Probably should confirm with @ajrobbins, but I think the spinner they wanted highlighted was not the one you did (previously black on white, now yellow on white, so now harder to see), but the smaller one above it, which is dark grey on light grey:

image

rajadain:
Here's what the current "Calculating Results" spinner looks like:

image

mmcfarland:
It may also be worth pointing out that the MMW equivalent that this is a descendant of has the spinner and text stacked and centered - possible a preferable look?

screenshot from 2017-06-29 10 15 51

rajadain:
I can still see the up/down buttons in Firefox

image

I correctly do not see them in Chrome, Safari, and IE11.

mmcfarland:
> I can still see the up/down buttons in Firefox

It may be easiest to just remove this by changing the `input` type or removing `step` from the control:

https://github.com/project-icp/bee-pollinator-app/blob/f608d55eeaa7661fce7742942efde1e3b75bd562/src/icp/js/src/modeling/templates/controls/userInput.html#L3-L9

rajadain:
Other aspects are working well. Failing tests are because of #240, so can be ignored here. I'll take another look once the reported issues have been addressed.

jfrankl:
Thanks for feedback. Please take another look @rajadain.

screen shot 2017-06-30 at 3 34 37 pm

image

rajadain:
Taking another look.

PR #243, opened 2017-07-07
Update supported domains; remove unused Nginx cache

hectcastro:
Ensure that pollinationmapper.org domains are used for all environments. Also, update the `DEFAULT_EMAIL_FROM` address so that the pollinationmapper.org domain is used.

Lastly, disable the unused Nginx OBSERVATION cache.

---

**Testing**

Navigate to the [new staging domain](https://staging.app.pollinationmapper.org/) and ensure that:

- The site works
- The certificate is valid
- The registration email comes from an pollinationmapper.org address

### DIFF
--- a/deployment/ansible/roles/bee-pollinator.app/templates/nginx-app.conf.j2
+++ b/deployment/ansible/roles/bee-pollinator.app/templates/nginx-app.conf.j2
@@ -1,10 +1,3 @@
-# Protects against scenarios where DNS records for the
-# site being proxied change.
-resolver 8.8.8.8 valid=300s;
-resolver_timeout 10s;
-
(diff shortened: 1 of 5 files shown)


deployment/ansible/roles/bee-pollinator.app/templates/nginx-app.conf.j2
```diff
@@ -1,10 +1,3 @@
-# Protects against scenarios where DNS records for the
-# site being proxied change.
-resolver 8.8.8.8 valid=300s;
-resolver_timeout 10s;
-
-proxy_cache_path {{ nginx_cache_dir }} levels=1:2 keys_zone=OBSERVATION:10m max_size=100m;
```
hectcastro: Yeah, good catch.

Issue #217, opened 2017-05-09
Add site tour


ajrobbins:
After evaluation, we've decided to use one of these two site tour libraries:
http://linkedin.github.io/hopscotch/
http://github.hubspot.com/shepherd/docs/welcome/

Jeff and I have a slight preference for hopscotch, because we don't want an overlay over the other site elements when one is highlighted for the tour. Pick the one that looks like the easiest to implement, however.

Here is a Marvel link with the site tour points and text (Click the comments button in the bottom left to see): https://marvelapp.com/h3g8796/screen/28035487

arottersman:
@ajrobbins/@jfrankl The tour spans three views: the initial crop map model page, the model page on the new scenario tab, and the compare view. When you click "next" on the last section of the initial crop map part of the tour, should it automatically take you to the new scenario tab? Or should the tour wait until the user has navigated of their own accord?

jfrankl:
The tour should wait until the user has navigated to the next view.

@ajrobbins could you please confirm that this is what we discussed?

ajrobbins:
@jfrankl @arottersman yes, I believe this is what we discussed and it makes sense to me.

PR #244, opened 2017-07-08
Add new user tour

arottersman:
## Overview

Add a tour to show new users around the model/compare stages of the app. 

We're using the tour library [hopscotch](http://linkedin.github.io/hopscotch/). It has its own API to handle state of seen/not seen using session storage, but it has several problems:
 - it only expects to be managing a single tour, and the tour for this app has three stages: when you're on the current crop map tab, when you're on "new scenario" and when you're in the compare view. These three stages were best implemented using three separate tours
- it resets the state after a tour has ended so you can't tell if you've never seen a tour or if you've completed one

Because of this I've mostly ignored what hopscotch thinks the state is and wrapped each tour object in our own object that manages state via local storage (and whether there's a user logged in with more than one project). 

See wireframes for desired text and placement https://marvelapp.com/h3g8796/screen/28035773

@jfrankl may also want a look at this 

Connects #217 

## Demo
screen shot 2017-07-07 at 8 13 06 pm

## Testing
- Pull branch and `./scripts/npm.sh install` to get hopscotch, then `./scripts/bundle.sh --vendor`
- As a guest user, progress to the model, confirm you see the initial tour bubble over current crop map
- Go through part of the tour, and then refresh the page or navigate back to draw
- Confirm the tour is right where you left it when you return
- Complete the tour and confirm it no longer appears on refresh
- Enter `localStorage.clear()` in the js console (the tour should reappear on subsequent refreshes)
- Login to an account with more than one project and confirm the tour doesn't appear
- Click around in any other order you can think ...

### DIFF
--- a/src/icp/bundle.sh
+++ b/src/icp/bundle.sh
@@ -15,6 +15,9 @@ STATIC_CSS_DIR="${DJANGO_STATIC_ROOT}css/"
 STATIC_IMAGES_DIR="${DJANGO_STATIC_ROOT}images/"
 STATIC_FONTS_DIR="${DJANGO_STATIC_ROOT}fonts/"
 
+# The hopscotch css src looks for images in "../img"
+HOPSCOTCH_IMAGES_SYMLINK="${DJANGO_STATIC_ROOT}img"
+
 BROWSERIFY="$BIN/browserify"
 ENTRY_JS_FILES="./js/src/main.js"
 
@@ -85,6 +88,7 @@ COPY_IMAGES_COMMAND="cp -r \
     ./img/* \
     ./node_modules/leaflet/dist/images/* \
     ./node_modules/leaflet-draw/dist/images/* \
+    ./node_modules/hopscotch/dist/img/* \
     $STATIC_IMAGES_DIR"
 
 COPY_FONTS_COMMAND="cp -r \
@@ -100,8 +104,11 @@ CONCAT_VENDOR_CSS_COMMAND="cat \
     ./node_modules/leaflet-draw/dist/leaflet.draw.css \
     ./node_modules/font-awesome/css/font-awesome.min.css \
     ./node_modules/bootstrap-table/dist/bootstrap-table.min.css \
+    ./node_modules/hopscotch/dist/css/hopscotch.min.css \
     ./css/shim/nv.d3.min.css \
     > $VENDOR_CSS_FILE"
+
+SYMLINK_HOPSCOTCH_IMG_DIR_COMMAND="ln -s $STATIC_IMAGES_DIR $HOPSCOTCH_IMAGES_SYMLINK"
 
 JS_DEPS=(backbone
          backbone.marionette
@@ -112,6 +119,7 @@ JS_DEPS=(backbone
          d3
          ./js/shim/marionette.transition-region.js
          ./js/shim/nv.d3.js
+         hopscotch
          jquery
          leaflet
          leaflet-draw
@@ -152,6 +160,7 @@ if [ -n "$BUILD_VENDOR_BUNDLE" ]; then
         $COPY_FONTS_COMMAND &
         $COPY_ZEROCLIPBOARD_COMMAND &
         $CONCAT_VENDOR_CSS_COMMAND &
+        $SYMLINK_HOPSCOTCH_IMG_DIR_COMMAND &
         $BROWSERIFY $BROWSERIFY_REQ \
             -o ${STATIC_JS_DIR}vendor.js $EXTRA_ARGS &
         $BROWSERIFY $BROWSERIFY_REQ $BROWSERIFY_TEST_REQ \
--- a/src/icp/js/src/core/createTour.js
+++ b/src/icp/js/src/core/createTour.js
@@ -67,5 +67,6 @@ module.exports =  function createTour(tour) {
     tourObject.onStart = _.bind(tourObject.initTour, tourObject);
     tourObject.onNext = _.bind(tourObject.incrementStep, tourObject);
     tourObject.onEnd = _.bind(tourObject.finishTour, tourObject);
+    tourObject.onClose = _.bind(tourObject.finishTour, tourObject);
     return tourObject;
 };
--- a/src/icp/js/src/core/createTour.js
+++ b/src/icp/js/src/core/createTour.js
@@ -61,7 +61,7 @@ var Tour = {
 };
 
 module.exports =  function createTour(tour) {
-    var tourObject = Object.assign(Object.create(Tour), tour);
+    var tourObject = _.assign(Object.create(Tour), tour);
 
     // Make sure the hopscotch callbacks bind to the tour object
     tourObject.onStart = _.bind(tourObject.initTour, tourObject);
--- a/src/icp/js/src/core/createTour.js
+++ b/src/icp/js/src/core/createTour.js
@@ -1,0 +1,71 @@
+"use strict";
+
+var _ = require('lodash'),
+    hopscotch = require('hopscotch'),
+    App = require('../app'),
+    modelingModels = require('../modeling/models');
+
+var Tour = {
+    getStatus: function() {
+        var step = localStorage.getItem(this.id);
+        return step !== null ? parseInt(step) : null;
+    },
+
+    shouldResume: function() {
+        return localStorage.getItem(this.id) < this.steps.length;
+    },
+
+    showTour: function() {
+        if (hopscotch.getState()) {
+            hopscotch.endTour();
+        }
+
+        var step = this.getStatus();
+        if (step === null) {
+            hopscotch.startTour(this);
+        } else if (this.shouldResume()) {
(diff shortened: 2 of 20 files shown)


rajadain:
Taking a look now.

rajadain:
The order of the prompts is different from the one specified in the wireframes here: https://marvelapp.com/h3g8796/screen/28035765

rajadain:
Ah nevermind, just saw the final comment.

src/icp/js/src/core/createTour.js
```diff
@@ -0,0 +1,73 @@
...
+
+    showTourIfNecessary: function() {
+        if (App.user.get('guest')) {
+            this.showTour();
+        } else {
+            var projects = new modelingModels.ProjectCollection();
+            projects
+                .fetch()
+                .done(_.bind(function() {
+                    var userHasProjects = !App.user.get('guest') && projects.length > 1;
```
rajadain: Do we need to check for `!App.user.get('guest')` again here? It has already been established by the parent `if` clause above.

src/icp/js/src/modeling/templates/scenarioTabPanel.html
```diff
@@ -1,4 +1,8 @@
-<a href="#{{ cid }}" data-scenario-cid="{{ cid }}" aria-controls="home" role="tab" data-toggle="tab" class="tab-name"><span>{{ name }}</span></a>
+<a href="#{{ cid }}" data-scenario-cid="{{ cid }}"
+    {% if is_current_conditions -%} id="current-crop-map-tab"
+    {% elif name == "New Scenario" -%} id="new-scenario-tab"
+    {% endif %}
```
rajadain: Should this be `{%- endif %}` to strip the final bit of whitespace?

src/icp/js/src/modeling/views.js
```diff
@@ -322,6 +323,15 @@ var ScenarioTabPanelView = Marionette.ItemView.extend({
         this.$el.toggleClass('active', this.model.get('active'));
     },
 
+    onShow: function() {
+        if (!this.model.get('active')) {
+            return;
+        }
+        var tour = this.model.get('is_current_conditions') ? tours.cropMapTour :
```
rajadain: These are always tricky to format. This could potentially read better:

```js
var tour = this.model.get('is_current_conditions') ?
           tours.cropMapTour : tours.newScenarioTour;
```

src/icp/sass/base/_tour.scss
```diff
@@ -0,0 +1,55 @@
+button.hopscotch-nav-button.next.hopscotch-next {
+    @extend .btn;
+    @extend .btn-active;
+    @extend .btn-md;
```
rajadain: I like that we're extending existing styles, but the button still looks a little tight:

image

A slight adjustment can make it look a little more balanced:

image

I changed padding of `hopscotch-bubble-container` from `15px` to `0.66rem`, and that of `hopscotch-nav-button next hopscotch-next` from `0 1rem` to `1rem`.

This can also be deferred to later if a design pass is upcoming.

src/icp/js/src/core/createTour.js
```diff
@@ -0,0 +1,73 @@
+"use strict";
+
+var _ = require('lodash'),
+    hopscotch = require('hopscotch'),
+    App = require('../app'),
+    modelingModels = require('../modeling/models');
```
rajadain: While I like the spirit and reasoning behind this, importing something from `modeling/` into `core/` feels a little smelly. However, we haven't introduced any new circular dependencies here, and I can't think of a better way of doing this, so I think we can let it slide for now. But since getting a list of user projects is used in many pages outside of `modeling/`, in the future we should consider moving it out into a less coupled module.

rajadain:
The tour's `z-index` should be less than any modal's:

image

src/icp/bundle.sh
```diff
@@ -100,9 +104,12 @@ CONCAT_VENDOR_CSS_COMMAND="cat \
     ./node_modules/leaflet-draw/dist/leaflet.draw.css \
     ./node_modules/font-awesome/css/font-awesome.min.css \
     ./node_modules/bootstrap-table/dist/bootstrap-table.min.css \
+    ./node_modules/hopscotch/dist/css/hopscotch.min.css \
     ./css/shim/nv.d3.min.css \
     > $VENDOR_CSS_FILE"
 
+SYMLINK_HOPSCOTCH_IMG_DIR_COMMAND="ln -s $STATIC_IMAGES_DIR $HOPSCOTCH_IMAGES_SYMLINK"
```
rajadain: This is also less than ideal, but not sure what better could be done.

src/icp/bundle.sh
```diff
@@ -152,6 +160,7 @@ if [ -n "$BUILD_VENDOR_BUNDLE" ]; then
         $COPY_FONTS_COMMAND &
         $COPY_ZEROCLIPBOARD_COMMAND &
         $CONCAT_VENDOR_CSS_COMMAND &
+        $SYMLINK_HOPSCOTCH_IMG_DIR_COMMAND &
```
rajadain: Thankfully running the `--vendor` command again and again does not cause issue, as this symlink is overwritten each time without complaint.

src/icp/js/src/modeling/tours.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+);
+
+var newScenarioTour = createTour(
+    {
+        id: "new-scenario-tour",
+        steps: [
+            {
+                target: "add-pollinator-planting-btn",
+                content: "Select a pollinator planting type from this menu, then draw it on the scenario map. Plantings can be drawn outside of the selected farm or field area, but cannot be stacked on top of each other. The minimum planting size is 100 ft x 100 ft. See help guide for more information.",
```
arottersman: Great idea. (It's this http://icpbees.org/tools-for-growers/help/)

src/icp/js/src/core/createTour.js
arottersman: Yeah, I feel that. `Projects` seem top level enough that they could be part of core (or we could bulk up `/projects`...though the project models are very tightly coupled with the modeling scenarios.... 🤔

src/icp/bundle.sh
arottersman: For reference here's the trouble https://github.com/linkedin/hopscotch/blob/dc747f0e7fc8009227810ed623e3d31e418cda9b/src/less/vars.less#L10

We could consider forking or shimming (would love to maintain `azavea/tour-bb` and show `linkedin/hopscotch` what's up!), but the symlink seemed the like the best patch for now.

arottersman:
Thanks, @rajadain! The two fixups address your comments. Here's an updated screenshot:

screen shot 2017-07-11 at 9 15 49 am

rajadain:
Taking another look at this.

arottersman:
While @ajrobbins was demoing this she noticed closing a tour via the "x" doesn't finish the tour — it will reappear if you come back to the page. I added c2fb35 to fix (adds a simple hook into the hopscotch `onClose` callback to finish the tour.) Mind testing this one last time, @rajadain?

- Start the tour current crop map tour
- Click a bubble's "x"
- Go to new scenarios
- Come back to current crop map
- The tour shouldn't reappear

ajrobbins:
Requests from the client:
1. Add a tour element for the "Go Back" button on Compare view, after the existing tool steps. Should read "Click here to return to the modeling view."
2. Add a tour element on the New Scenario page, after the existing steps, for the Login button. Should read "Login or create an account to save this project."

rajadain:
Let me know when this is ready for another look.

arottersman:
Oh, boy, is it ready!

rajadain:
[Taking another look](https://youtu.be/txw0tYo13jw)

rajadain:
I suppose this could go either way, but I feel like context menu popups should be above the tour popups:

image

Also, can we make the distance and duration of the transitions a little shorter? About 75% of the current length / time would be better.

Other than that, I tested the additional popups, and the link to the help section. Everything works as expected. Great work!

rajadain:
https://github.com/project-icp/bee-pollinator-app/issues/250 may be caused by this branch

rajadain:
The `_.assign` fix now has this working again in IE11. IE11 doesn't seem to render shadows though, which makes the popup arrow look cutoff in some cases:

image

This can be alleviated by adding a 1px dark border to the whole thing, either in all cases or just for IE.

rajadain:
We may also consider trying to increase the shadow's visibility in IE11, as per discussion here https://stackoverflow.com/a/22517756/6995854

ajrobbins:
For now, let's treat the shadow as an enhancement. Turbo Pteradactyl

arottersman:
Adjusted the z-index and animation time. Good to merge? Will make a card for the shadow.

rajadain:
Checking out the final bits

arottersman:
Thanks for all your reviewing on this!

Issue #150, opened 2017-02-03
Fix scenario renaming


arottersman:
When renaming a scenario:
- the cursor begins on the lefthand side
- when you try to select text by clicking it, rename mode exits and you end up just clicking the tab
- when you press enter but don't change anything you get warned that there's already a scenario with that name
- you can rename to a blank string

Could bring over from MMW the work done for https://github.com/WikiWatershed/model-my-watershed/pull/1683 .

arottersman:
@ajrobbins @jfrankl Should the update mimic MMW prime where the tab text becomes editable, or should it mimic the change made most recently (https://github.com/WikiWatershed/model-my-watershed/pull/1984) and create a rename modal?

ajrobbins:
The former - tab text becomes editable

PR #245, opened 2017-07-11
Scenario rename behavior enhancements

arottersman:
## Overview

- add additional unit tests for renaming scenario behavior
- allow renaming to same name as the scenario already had
- disallow renaming to blank name
- allow double click on text to highlight
- start with all text highlighted

(This PR is basically identical to WikiWatershed/model-my-watershed#1683)

Connects #150 

## Demo
lg3uax1xb6

## Testing instructions
- Rename scenarios confirming they begin highlighted, can be double clicked, and cannot be named another scenario's name or an empty string


arottersman:
Thanks!

Issue #254, opened 2017-07-12
Add beta symbol to site name in nav bar


ajrobbins:
The ICP team would like to emphasize that the Pollination Mapper is a beta product in terms of model output, so users don't write off the tool if they see model results that don't match their expectations. Add a beta symbol to the site name in the nav bar to remind users of this.

ajrobbins:
Shares a point with #248

ajrobbins:
![Pasted image at 2017_07_13 03_23 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5685332b20438e130091808c/125516c66e82c728ace21e0d46d89589378787e986e4be44a11df180691477227e3bf37c4fd4b068ef1144535c4f2ed153444d04bea6d37ef5f22e7a98590ebf80655a73c33d17a8b9d44dbb05ba208f1b83e7efdf5edcfa7f74b9dfd0126754505a9bf54f37a6934acc51ff42e64b57ac153e65e4647b.png)

Issue #251, opened 2017-07-12
Yellow text should be teal in Compare view modification list


ajrobbins:
In the Compare view mods list, the area figure is still in the yellow which we changed to a teal color elsewhere in the app in https://github.com/project-icp/bee-pollinator-app/pull/238:

![screen shot 2017-07-12 at 10 43 35 am](https://user-images.githubusercontent.com/12868342/28123506-1b3155ba-66ef-11e7-9251-1ace8021ca3c.png)

arottersman:
The teal color is actually `Yuba Blue` https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/sass/utils/_variables.scss#L25

arottersman:
Shares a point with #225

Issue #249, opened 2017-07-12
Update "help page" link on landing page


ajrobbins:
Help page link currently opens in same tab:

![screen shot 2017-07-12 at 10 23 49 am](https://user-images.githubusercontent.com/12868342/28122467-46baa1bc-66ec-11e7-9426-438af593d272.png)

Should open in new tab.

Also, help link URL will be changing before the release, likely to a subdirectory of pollinationmapper.org.

arottersman:
Shares a point with #225

ajrobbins:
Just checked with ICP - we can just use pollinationmapper.org to replace the ICP page help links.

Issue #225, opened 2017-05-16
Reduce # of decimal places in yield table


ajrobbins:
The current yield table shows figures with three decimal places of specificity. Reduce to one for simplicity.

arottersman:
Shares a point with #249 and #251

PR #265, opened 2017-07-17
Small UI changes

mmcfarland:
## Overview

Some small fixups requested by the client:
* Number are uniformly rounded to a single decimal place.
* Yuba Blue is used for output text color on compare
* "BETA" is added to the header branding.
* Updates helpguide link to new window & pollinationmapper.org

Connects #254 
Connects #225 
Connects #251 
Connects #249 

## Notes
~~I matched the header style from the issue for the beta text, but also tried a version as a <sup>superscript</sup> as is common.  Looks like the following, we can do that if consensus indicates it looks better, I do.  I also did not add it to the tab title since we're not changing it to say "beta" every place within the app that it appears, let me know if you think otherwise.~~

*Update: we decided to user the superscript text as represented in Demo*

## Demo

**Rounding and color**
screenshot from 2017-07-17 10 57 10
screenshot from 2017-07-17 10 50 13
screenshot from 2017-07-17 10 49 29

**Header text**
screenshot from 2017-07-17 11 03 57

### Testing
Verify that all numeric output text is rounded to a decimal place and is the appropriate color.  This includes:
* Shared and scenario popups
* Modification dropdown
* Compare output
* Yield model output

Check that `BETA` is in the header, as well.

### DIFF
--- a/src/icp/js/src/core/templates/header.html
+++ b/src/icp/js/src/core/templates/header.html
@@ -1,5 +1,5 @@
 <nav id="app-header" class="navbar-watershed">
-    <div class="brand"><a href="/">Pollination Mapper</a></div>
+    <div class="brand"><a href="/">Pollination Mapper <sup>BETA</sup></a></div>
     <div class="navigation">
         <ul class="main">
             {% if guest %}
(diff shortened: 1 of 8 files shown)


ajrobbins:
I like the superscript, and agree the change should only be to the nav bar header, not everywhere.

ajrobbins:
If you want to toss https://github.com/project-icp/bee-pollinator-app/issues/249 in, just change the help link to point to pollinationmapper.org.

arottersman:
Heads up there's also a link to the help page in the tour 
https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/js/src/modeling/tours.js#L45

mmcfarland:
Updated the PR description and commits to use the superscript version of beta text and also adjust the help guide links.

Issue #273, opened 2017-07-31
Crop names overlap in compare view chart


ajrobbins:
Observed on 13" macbook, Chrome:

Ocurrs when more than 3 crops are present (or added via land cover mods):
![screen shot 2017-07-28 at 5 16 53 pm](https://user-images.githubusercontent.com/12868342/28780005-39c08d3c-75d3-11e7-9595-576d67216380.png)

With only three:
![screen shot 2017-07-28 at 5 18 17 pm](https://user-images.githubusercontent.com/12868342/28780272-23c02fa0-75d4-11e7-8bdd-dced7362effd.png)

Perhaps there's a way to hyphenate and wrap them, or have them tilting down to save space? Another option would be to force the column to be wider.

PR #275, opened 2017-07-31
Rotate crop chart label to prevent crowding

mmcfarland:
## Overview

When the number of crops is above 2 on compare, or 3 on the main
output, rotate the crop name slightly to avoid squished text. The
labels are not rotated under all circumstances because they look off
when there is plenty of room but are still rotated.

Connects #273 

### Main output, changes at 3 crops
screenshot from 2017-07-31 10 23 02
screenshot from 2017-07-31 10 22 31

### Compare output, changes at 2 crops
screenshot from 2017-07-31 10 19 53
screenshot from 2017-07-31 10 18 52

### DIFF
--- a/src/icp/js/src/core/chart.js
+++ b/src/icp/js/src/core/chart.js
@@ -135,13 +135,15 @@ function renderGroupedVerticalBarChart(chartEl, data, options) {
 
     options = options || {};
     _.defaults(options, {
-        margin: {top: 20, right: 30, bottom: 40, left: 60},
+        margin: {top: 20, right: 30, bottom: 45, left: 60},
         maxBarWidth: 150,
         showLegend: true,
         disableToggle: true,
     });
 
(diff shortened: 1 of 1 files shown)


arottersman:
Taking a look now

src/icp/js/src/core/chart.js
```diff
@@ -152,6 +154,11 @@ function renderGroupedVerticalBarChart(chartEl, data, options) {
             addSvgClass($svg, 'compare');
         }
 
+        // Rotate the labels to conserve space when there are X+ items
+        if (options.barClasses.length > cropCountBeforeRotate) {
+            chart.rotateLabels(18);
```
arottersman: I think I'm more used to seeing labels rotated the other way, ie `-18`, but I don't think it matters.
screen shot 2017-08-01 at 11 00 06 am

Issue #284, opened 2018-09-06
Discuss Bees Front-End 2.0


rajadain:

fungjj92:
Draft at https://github.com/project-icp/bee-pollinator-app/blob/jf/phase2-adr/doc/Phase%202/000_project_architecture_adr.md @rajadain 

While we talked about it in the meeting, I excluded discussion of how versions of client uploaded data would be signified to map click entries because that discussion seemed premature for now.

rajadain:
Thanks for putting that together! Since I can't comment on the file directly, I'm posting some initial thoughts here:

---

> There will be dedicated user accounts as well as the ability to sign in with credentials from app.pollinationmapper.org.

My impression was that each app will have its own user accounts, and they will be separate. If a user already has an account in 1, they'll have to make a new one for 2. We should confirm that with @mmcfarland.

> where will 2 live?

"live" is ambiguous here, as it refers both to the deployment environment and the repo. It is valuable to differentiate the two and talk about their separation from 1's repo and deployment individually.

> If separate (opposite, for combined)

Consider making this into a table. That would be easier to read than flipping a list in one's mind.

> The crux of the decision comes down to auth.

Consider moving this paragraph to the previous section. By the time we get to the "Decision" part of the ADR, all the influencing factors should have been discussed.

> syncing user tables

This was never an issue, since 2 isn't going to have its own back-end or its own user tables. Whether we developed 2 as a separate front-end project or a sibling Django template, we were always going to use the existing user table for user management. The one tweak we'll need to make to the existing table (as mentioned in the planning meeting) is to add a column to the user table which indicates to which app the user signed up from.

> Bonus: reduced operations labor.

While it is true that 2 and API will be deployed together, this also means that 2, which will be a simple static front-end app, and significantly simpler to deploy on its own, cannot be deployed without the overhead of also deploying the API.

> there is a certain amount of uniqueness we want to update the API with as we build out 2

I'm not following this very well. By "uniqueness" do you mean custom functions that will be used only in 2 and not 1? Consider rephrasing "certain amount of uniqueness" to "set of custom functionality".

fungjj92:
Thanks for looking promptly. Oof, let me make the branch into a PR so it is easier to work with. I'll copy over your comments.

PR #285, opened 2018-09-19
Phase 2 ADR

fungjj92:
I drafted meeting notes from Monday 9/17 as an ADR about the phase 2 bees project architecture.
Handing off to Terence for review for now. Will invite wider feedback when the ADR is in a later state draft.

Connects #284


doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
+# Project Architecture ADR
+
+The Phase 2 of the Bee Pollinator Project (2) will include a new Javascript
+(React/Redux) front-end communicating with the existing [Phase 1 Bees
+API](https://app.pollinationmapper.org/) (1 or API). There will be dedicated
+user accounts as well as the ability to sign in with credentials from
```
fungjj92: > My impression was that each app will have its own user accounts, and they will be separate. If a user already has an account in 1, they'll have to make a new one for 2. We should confirm that with @mmcfarland.
- @rajadain 

Ooh, this looks like a fundamental misunderstanding pretty crucial to this ADR. Clarification solicited, indeed.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+https://app.pollinationmapper.org/.
+
+While we are waiting for final
+[wireframes](https://app.goabstract.com/projects/1955fff0-89e6-11e8-9d27-3b6b7c64f4e5/branches/master/files/E258E310-7858-4F64-9C6F-43572DBBEB19),
+client feedback, and a contract, we don't expect breaking changes at the project
+architecture level and are moving ahead planning the project's setup.
+
+## Context
+
+The main question driving this discussion is -- where will 2 live? The options
```
fungjj92: > "live" is ambiguous here, as it refers both to the deployment environment and the repo. It is valuable to differentiate the two and talk about their separation from 1's repo and deployment individually.

from @rajadain 

Point taken. I'm really referring to where the repos will be: I'll clarify.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+have implications on hosting, CORS, user management, deployment, etc. to be
+discussed here.
+
+### To separate or combine projects?
+The only two options really are will 2 live in its own repo and make CORS
+requests to the API or will it live within the Phase 1 repo.
+
+#### If separate (opposite, for combined):
```
fungjj92: > Consider making this into a table. That would be easier to read than flipping a list in one's mind.
from @rajadain

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+    - Implicates 2 having separate users
+    + Easier to navigate project during development
+    + Prevent linked deployments
+
+## Decision
+
+After discussion, we decided on combining the two projects whereby 2's project
+code will be nested in [1](https://github.com/project-icp/bee-pollinator-app).
+
+The crux of the decision comes down to auth. We want users of 2 to be able to
```
fungjj92: > Consider moving this paragraph to the previous section. By the time we get to the "Decision" part of the ADR, all the influencing factors should have been discussed.
from @rajadain

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+create accounts on 2 or sign in with their API user accounts. We want access to
+the API user data table and Django auth and session management and combining
+projects avoids unwanted complexity of setting up a cross-site token auth
+system, syncing user tables, or other potential solutions for a separate
```
fungjj92: > This was never an issue, since 2 isn't going to have its own back-end or its own user tables. Whether we developed 2 as a separate front-end project or a sibling Django template, we were always going to use the existing user table for user management. The one tweak we'll need to make to the existing table (as mentioned in the planning meeting) is to add a column to the user table which indicates to which app the user signed up from.
from @rajadain

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+projects.
+
+
+## Consequences
+
+Combining project directories plays nicely in other regards, too. Combined, 2
+will be deployed along with 1 and hosted on the same EC2 instances. Ultimately
+we don't expect high usage of either app such that overloaded instances or even
+resource scaling becomes an issue. Bonus:  reduced operations labor.
```
fungjj92: > While it is true that 2 and API will be deployed together, this also means that 2, which will be a simple static front-end app, and significantly simpler to deploy on its own, cannot be deployed without the overhead of also deploying the API.
from @rajadain

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,65 @@
...
+
+From the perspective of a developer, noteworthy wins to building on the API are
+avoiding cobbling together workarounds to handle CORS requests to the API, API
+user authentication, and API user sessions. Devs will only have 1 repo's dev
+environment to run, albeit many VMs, which is nice too.
+
+Thinking ahead, there is a certain amount of uniqueness we want to update the
```
fungjj92: > I'm not following this very well. By "uniqueness" do you mean custom functions that will be used only in 2 and not 1? Consider rephrasing "certain amount of uniqueness" to "set of custom functionality".
from @rajadain

fungjj92:
Redrafted and pushed. 

I realized I was confused about and missing some detail that clarification and rewriting have sunken in my head a lot better. There's a lot more detail now in the Context that builds a stronger Decision, and I thusly I had a better idea of how to imagine the Consequences. Are there other major details we/you had thought about for the Context that are missing from my list? Any logical discrepancies?

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
+# Project Architecture ADR
+
+Phase II of the Pollination Project (Beekeeper) is a web app designed for
+beekeepers to be able to analyze site conditions for apiaries. They do so by
+clicking on a map, which will kick raster analysis jobs, and return said data to
```
rajadain: > return said data

The data hasn't been "said" so far. Consider adding a line about what the data is for (e.g. data for identifying optimum positioning of apiaries).

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+the user's browser. Previously it was decided that the app will be a front-end
+app (React) that makes requests against the [Phase 1 Bees
+API](https://app.pollinationmapper.org/) (Pollination Mapper or API) and this
+ADR takes the next steps and should provide a way forward.
```
rajadain: > should provide a way forward

Let's be more confident here 😄 Consider rephrasing to "provides a way forward"

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+While we are waiting for final
+[wireframes](https://app.goabstract.com/projects/1955fff0-89e6-11e8-9d27-3b6b7c64f4e5/branches/master/files/E258E310-7858-4F64-9C6F-43572DBBEB19),
+client feedback, and a contract, we don't expect breaking changes at the project
+architecture level and are moving ahead planning the project's setup.
+
+## Context
+
+Some key questions drive this discussion. 
+
+Where will the code live? How will Beekeeper's requests to the API be authorized
```
rajadain: Consider listing these questions, rather than inlining them in a paragraph.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+the Pollinaton Mapper repo. Below thinks through the implications of each.
+
+- Project syncrony -- We can anticipate actively customizing the API for
+  Beekeeper. In order for Beekeeper's staging and production sites to function,
+  both projects need to be in sync with one another. This is a tricky
+  relatonship to maintain as it will require more laborious, linked deployments
+  as well as being careful not to break Beekeeper.
+
+- Hosting -- As a front-end only static app, Beekeeper has many potential
+  hosting options such as Netlify, Heroku, AWS s3, or Azure (trollface). How we
```
rajadain: > (trollface)

Consider removing. A day may well come when we have to host on Azure.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+      potentially by devs, and not even operations themselves. Of course, this
+      set up comes with external considerations discussed below. 
+
+    - CORS -- Should Beekeeper be hosted on its own domain, such as would
+      provide Netlify or Heroku, it will run into CORS in the browser. The API
+      is not CORS-enabled and Beekeeper as an outside domain making requests
+      would trigger CORS errors. Enabling CORS and whitelisting Beekeeper as a
+      valid hostname is possible, albeit opens up vulnerability.
+
+    The API is hosted on an EC2 instance with its own dedicated IP. Similarly to
```
rajadain: Is this paragraph its own or a continuation of the previous CORS bullet point? If the latter, indent one level deeper to maintain correct indentation in both the raw and rendered Markdown.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+    the MMW/BigCZ project, we could circumvent CORS by implementing [name-based
+    shared web
+    hosting](https://en.wikipedia.org/wiki/Shared_web_hosting_service) whereby
+    the projects share an IP and the existing load balancer maps the request to
+    the correct website based on its request header. This would resolve a need
+    for CORS relaxing.
+
+    - CSRF and session auth - The API's endpoints are protected by Django's CRSF
```
rajadain: > CRSF

Typo: CSRF, not CRSF. It stands for Cross-Site Request Forgery.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+      middleware. Beekeeper will need to have an authorized Django user session
+      and CSRF token to make HTTP requests to the API beyond GET requests, which
+      are exempted. If the apps are separate we will have to open Beekeeper
+      endpoints to be CRSF-exempt and session-exempt. This is a security risk
```
rajadain: > CRSF-exempt

here too.

doc/Phase 2/000_project_architecture_adr.md
fungjj92: Reformat: Oops, this is supposed to be under CORS

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+Some key questions drive this discussion. 
+
+Where will the code live? How will Beekeeper's requests to the API be authorized
+and authenticated? Besides using the API, how else will Beekeeper and
+Pollination Mapper be connected?
+
+As to where the code will live, the 2 options are its own repo, or built into
+the Pollinaton Mapper repo. Below thinks through the implications of each.
+
+- Project syncrony -- We can anticipate actively customizing the API for
```
fungjj92: Bold the sections and typo syncrony --> synchronicity

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+## Decision
+
+Overall, combining the two projects whereby Beekeeper's code will be nested in
+[Pollination Mapper](https://github.com/project-icp/bee-pollinator-app) will
+provide more tangible benefit by avoiding significant feature accomodations to
+access the API as well as maintaining project synchronicity. The potential
+benefit to separate projects we're giving up is the ease of static site
+deployment via a platform like Netlify, but I think synchronicity between
```
rajadain: > I think

As this is a project document, consider avoiding the first person. Consider replacing with something like: "but developer time saved with synchronicity between Beekeeper and the API is more valuable than the additional deployment time".

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,124 @@
...
+resource scaling (i.e. Celery servers) becomes an issue.
+
+
+## Consequences
+
+Looking ahead, we can create the project folders for Beekeeper and scaffold a
+working app in the [API](https://github.com/project-icp/bee-pollinator-app)
+repo.
+
+For the backend, we can create a sibling Django project and get started building
```
rajadain: Do we want to create an entire Django project (with its own settings and everything) or just an app in the existing Django setup? I thought the latter.

doc/Phase 2/000_project_architecture_adr.md
rajadain: Were all three questions answered?

How will the Beekeeper requests be authorized? Using CSRF?

Besides the API, how else are the two apps connected?

It's unclear if these questions are answered.

fungjj92: I added a little bit to the decision about authorization, but I think the other two questions are answered. I don't say in a sentence "The apps are connected by X, Y, Z", rather the ADR as a whole answers the q -- nested Django app, share the same codebase, duplicated custom endpoints, separate JS folder and bundle, hosting infra, etc.

fungjj92:
I think this should be ready for wider review ?

rajadain:
Yes. I'll take another look, but we should invite others too.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,130 @@
...
+ahead planning the project's setup.
+
+## Context
+
+Some key questions drive this discussion. 
+
+- Where will the code live?
+- How will Beekeeper's requests to the API be authorized and authenticated?
+- Besides using the API, how else will Beekeeper and Pollination Mapper be
+  connected?
```
kellyi: Consider making each of these questions a subsection, then elaborating a bit on the answer.

Currently it seems like the subheadings below address them all together, but it when this is re-read in the future it might be worthwhile to take the subheadings below and reorganize them as the elaboration of one of the above qs.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,130 @@
...
+  returned to the user. This is very similar to that of Pollination Mapper and
+  we can leverage the Celery sync job pipeline by combining projects.
+
+- **Developer experience** -- The developer will need to run the Beekeeper
+  project against a local instance of the API. It would be convenient to only
+  spin up one project, albeit multiple VMs, rather than two projects. We would
+  also have to deal with CORS and session auth between separate VMs, as
+  discussed above. Additionally, it is easier to work in just one code base and
+  may save valuable time. For example, the time saved to search for a file or
+  some bit of code that is in reality in the other repo.
```
kellyi: This is possibly for a future ADR but I'm curious how we imagine our usual Docker/webpack-dev-server/React setup working within this dev environment?

Are we planning to add a Docker container to run webpack-dev-server inside the app VM? Or are we planning to add another VM which runs the React app? (I think the latter would require making cross-origin requests, but so might the former?)

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Decision
+
```
kellyi: Consider also having three sections here: since the context started out with three questions the ADR would answer, seems like this should summarize those answers in a way future us can read at a glance.

doc/Phase 2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,130 @@
...
+able to distinguish between app codes, and also allow better tracking of API
+usage, if those metrics are so desired. Distinctive labelling should also apply
+to naming views, models, tests, etc. that live in the API.
+
+Lastly, there is already a JS front-end for Pollination Mapper. Its files are
+distributed between app root and a nested `js` folder. Because Beekeeper will
+have significant JS files and similar bundling pipeline as well, we will want
+all Beekeeper JS including the `package.json` and requirements in a nested
+folder a level down to avoid conflict. For better organization, consider
+repackaging all Pollination Mapper JS into a similarly nested folder as well.
```
kellyi: Are we going to do another ADR to determine how to set up and structure the React app?

Given the decision to keep these projects in the same repo and to run the new React client within the same domain as the Django app, it seems like it'd be worthwhile to discuss the logistics of setting up the React app for both development and production a bit more prior to starting that work.

If we're porting our standard React setup over, we'll have a Docker container for development running webpack-dev-server somewhere (that won't be same-origin). Will that live in the app VM or in another VM?

If we're adapting the React setup for this but not using webpack-dev-server/the Docker container, how will we serve the app in development? It'd be nice to have hot-module-reloading set up for the React app, but I'm not sure how that would work if it was just using `webpack --watch ....`.

If we don't use a container, are we going to use the version of Node used in the current app? Looks like the app is currently using `0.10.32` -- https://github.com/azavea/ansible-nodejs/blob/0.3.0/defaults/main.yml -- which is out of support and most likely not completely compatible with the Civic Apps App Template setup. 

I remember that I had to upgrade the Node version on Jenkins01 from 0.10 something to 4.x something to get the Garden Exemptions React application to build. We could possibly upgrade the Node version here to 8.x something, but it seems possible/probably that doing so would break some pieces of the existing Pollination Mapper client.

Similar questions about building for production: seems like part of the production build process would have to involve having webpack create a bundle and write it somewhere. Will that be done using a containerized version of Node or using the app's existing version of Node?

Part of the production build process for our React apps also involves writing out the index.html file from a template and injecting the app bundle's filename with it's hash identifier. Will that be compatible with keeping the existing index.html file for the Pollination Mapper, but adding a separate client depending on the URL? (I think that's how the MMW/BigCZ stuff worked.) Or will the build process use a separate index.html file template and write it somewhere else?

rajadain:
Thanks for pointing that out @kellyi. I met with @fungjj92 and we spoke of the alternatives available at this point. We don't know this for sure, but it is a very likely possibility that **React will not work with Node v0.10.32**, the [current version](https://github.com/azavea/ansible-nodejs/blob/92245ba10c25b2aef53a47bb3e3efb81334617c5/defaults/main.yml#L2) [installed](https://github.com/project-icp/bee-pollinator-app/blob/1cf3c26d54d76ee596c33e370b48bc3722f57378/deployment/ansible/roles.yml#L7-L8) in the Bees VM.

There are three ways to address this problem:

1. **Upgrade Node version in Bees VM**
    - This will allow us to use React
    - This could potentially break the UI for Pollination Mapper, and start a chain reaction requiring us to update a great number of project dependencies, not all of which may be compatible with each other

2. **Use Backbone instead of React**
    - This will allow us to keep the current Node version
    - Backbone is an older technology that is not actively maintained anymore (the latest version 1.3.3 is from two years ago), and we're on an even older version of Backbone and Marionette
    - Starting a new project with old technology is risky, especially as we may be maintaining this project for some time to come, during which the technology will age more and become harder to use

3. **Make a separate single-page app for Beekeepers**
    - This will allow us to use React and keep the current Node version of Pollination Mapper
    - The request authentication benefits (CSRF / CORS) we could get from colocating the new front-end with the API on the same domain will be lost. We'll have to figure out how to authenticate API calls (JWT? OAuth? Simple Token?), and support it both in the front- and back-ends.
    - Given the need to authenticate API calls, and the relative insecurity of allowing API access to a specific domain or hard-coded API key, both of which are available to the client, we may have to rely more on user authentication. This could move the need for users to login at a much earlier stage in the app workflow than previously thought.
    - While deploying a static app may be easier, we will have to ensure correspondence with the separately deployed API version

I think we should consult with @mmcfarland upon his return before making a final selection, and then update the ADR accordingly.

doc/Phase 2/000_project_architecture_adr.md
rajadain: After thinking about this some more, here's the answers I have right now:

> If we're porting our standard React setup over, we'll have a Docker container for development running webpack-dev-server somewhere (that won't be same-origin). Will that live in the app VM or in another VM?

I think it makes sense for this to live in the App VM. The front-end is in the App VM, and as part of this front-end, so should this.

> If we're adapting the React setup for this but not using webpack-dev-server/the Docker container, how will we serve the app in development?

I think initially we'll just use a script similar to `bundle` to build the dependencies. As Django will handle delivering the HTML for the front-end, it will always be up on :8000, so we don't need a separate `server` script.

> If we don't use a container ...

We're going to use a container.

> Similar questions about building for production: seems like part of the production build process would have to involve having webpack create a bundle and write it somewhere. Will that be done using a containerized version of Node or using the app's existing version of Node?

Containerized version.

> Part of the production build process for our React apps also involves writing out the index.html file from a template and injecting the app bundle's filename with its hash identifier. Will that be compatible with keeping the existing index.html file for the Pollination Mapper, but adding a separate client depending on the URL? (I think that's how the MMW/BigCZ stuff worked.) Or will the build process use a separate index.html file template and write it somewhere else?

The app URL is available when rendering the home view. We can use it to switch things out in the template, or use an entirely different template. I think we'll start by turning off adding hashes in webpack for now, and add it at a later stage.

rajadain:
Added the discussion from above, as well as an outline for next steps, to the ADR. This could probably take one final look before merging.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+makes requests against the [Phase 1 Bees API] (Pollination Mapper or API) and
+this ADR takes the next steps and provides a way forward.
+
+While we are waiting for final [wireframes], client feedback, and a contract, we
+don't expect breaking changes at the project architecture level and are moving
+ahead planning the project's setup.
+
+## Context
+
+Some key questions drive this discussion.
```
kellyi: Seems like we've sublated all of these questions in the course of discussing this and the new thing this ADR should record should be something like:

"Since we are going to create a new client frontend for Beekeeper, how are we going to set up the development environment for that new frontend?"

I think the 4 options enumerated below should be placed under an ` ## Options` heading and each subsection there should catalog lists of pros and cons to each option.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+      would trigger CORS errors. Enabling CORS and whitelisting Beekeeper as a
+      valid hostname is possible, albeit opens up vulnerability.
+
+      The API is hosted on an EC2 instance with its own dedicated IP. Similarly
+      to the MMW/BigCZ project, we could circumvent CORS by implementing
+      [name-based shared web hosting] whereby the projects share an IP and the
+      existing load balancer maps the request to the correct website based on
+      its request header. This would resolve a need for CORS relaxing.
+
+    - **CSRF and session auth** - The API's endpoints are protected by Django's
```
kellyi: Since these & the other existing bullet points are no longer open questions, consider fusing all of these summaries into a single paragraph.

kellyi: Also consider adding a specific sentence noting that the version of NodeJS in the existing app VM is 0.10.32 -- which is one of several large constraints on how we set up the new client application.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+      combined projects, we have direct access to the Django session and thusly
+      CSRF management.
+
+- **Resource needs** -- The workflow of the Beekeeper app will be quite similar
+  to the Pollination API. As described in the intro the app workflow will kick
+  off a likely async raster processing job where the numerical data will be
+  returned to the user. This is very similar to that of Pollination Mapper and
+  we can leverage the Celery sync job pipeline by combining projects.
+
+- **Developer experience** -- The developer will need to run the Beekeeper
```
kellyi: Consider cutting this item altogether. The explanation makes sense but it seems like the decision the ADR's now recording is about how to structure the new client within the same repository rather than whether to include the new client in the same repository.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+  project against a local instance of the API. It would be convenient to only
+  spin up one project, albeit multiple VMs, rather than two projects. We would
+  also have to deal with CORS and session auth between separate VMs, as
+  discussed above. Additionally, it is easier to work in just one code base and
+  may save valuable time. For example, the time saved to search for a file or
+  some bit of code that is in reality in the other repo.
+
+Furthermore, the [current version] of `node` [installed] in the App VM
+(0.10.32) will most likely _not_ run webpack or React. The possible
```
kellyi: In addition to webpack & React, our app template has a bunch of other dependencies that may not be compatible. Consider revising this to indicate that "it's not clear the app template's NPM dependencies are compatible with NodeJS 0.10.32" and note that setting up a project similar to our app template for Garden Exemptions required upgrading Node on Jenkins01 from 0.10.something to 4.something

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -77,6 +77,50 @@ the Pollinaton Mapper repo. Below thinks through the implications of each.
...
+    - While deploying a static app may be easier, we will have to ensure
+      correspondence with the separately deployed API version
+
+4. **Use a Docker for building React JavaScript assets**
+    - Just as we have a bundle script for Pollination Mapper, we use a bundle
+      script that runs a Docker container with a proper version of Node to
+      build the assets. This Docker container is only used at build time.
+    - This will allow us to use React and not touch the existing code for
+      Pollination Mapper
+    - We will not have hot-reloading, but we wouldn't have it in a Backbone
```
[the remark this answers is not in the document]
kellyi: I *think* @rajadain demoed a setup with hot reloading working yesterday, so we could possibly cut this line item altogether.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+      version either
+    - The build pipeline will look very similar to the existing one for
+      Pollination Mapper
+
+## Decision
+
+Overall, combining the two projects whereby Beekeeper's code will be nested in
```
kellyi: Consider cutting this paragraph and making the second paragraph below the focus of the decision.

doc/phase-2/000_project_architecture_adr.md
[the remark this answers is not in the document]
fungjj92: He did, but I think this ADR should still capture any uncertainty/gambles we had before pursuing a strategy

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -92,6 +136,19 @@ Ultimately we don't expect high usage of either app such that sharing a load
 balancer (AWS ELB) or even resource scaling (i.e. Celery servers) becomes an
 issue.
 
+We go with **Alternative 4** from above. The new front-end files will be stored
```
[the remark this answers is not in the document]
rajadain: I kept the fact that we're using Alternative 4 in "Decision", but moved the implementation details to "Consequences"

doc/phase-2/000_project_architecture_adr.md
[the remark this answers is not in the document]
rajadain: Rephrased to allude to challenges:

> Achieving hot-reloading may be challenging as we need to serve from Django and not WebPack Dev Server, but on the other hand Backbone would not have had any hot reloading at all.

doc/phase-2/000_project_architecture_adr.md
rajadain: The Node app version is specified later on.

The first set of factors have been grouped under **Architectural Factors**, and the second under **Implementation Factors**. I think keeping the original discussion is useful for retrospective investigation, because it is not recorded anywhere else.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,199 @@
...
+able to distinguish between app codes, and also allow better tracking of API
+usage, if those metrics are so desired. Distinctive labelling should also apply
+to naming views, models, tests, etc. that live in the API.
+
+Lastly, there is already a JS front-end for Pollination Mapper. Its files are
+distributed between app root and a nested `js` folder. Because Beekeeper will
+have significant JS files and similar bundling pipeline as well, we will want
+all Beekeeper JS including the `package.json` and requirements in a nested
+folder a level down to avoid conflict. For better organization, repackaging all
+Pollination Mapper JS into a similarly nested folder may be considered as well.
```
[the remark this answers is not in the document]
rajadain: Added structure of Dockerized Beekeeper setup. Also removed the line about changing existing structure.

rajadain:
If there are no further comments, I will merge this later today.

doc/phase-2/000_project_architecture_adr.md
```diff
@@ -0,0 +1,239 @@
...
+      Simple Token?), and support it both in the front- and back-ends.
+    - Given the need to authenticate API calls, and the relative insecurity of
+      allowing API access to a specific domain or hard-coded API key, both of
+      which are available to the client, we may have to rely more on user
+      authentication. This could move the need for users to login at a much
+      earlier stage in the app workflow than previously thought.
+    - While deploying a static app may be easier, we will have to ensure
+      correspondence with the separately deployed API version
+
+4. **Use a Docker for building React JavaScript assets**
```
fungjj92: typo -- "Use Docker" or "Use a docker container"

fungjj92:
I cant approve since I opened the PR. I reread and it LGTM!

PR #294, opened 2018-10-19
Update deployment pipeline

tnation14:
# Overview
This PR includes fixes for some of the inconsistencies I found when doing a Staging deployment. Changes are mostly version bumps and making sure dependencies are consistent across environments.

## Changes
- [ ]  Add missing Ansible dependencies to `roles.yml`
- [ ]  Make `libpq` and `psycopg2` versions consistent across environments
- [ ]  Lookup Ubuntu AMI IDs with `boto` instead of manually parsing Ubuntu's `releases.current.txt` file.
- [ ] Bump Ansible version to `2.5.4`

# Notes
-  The `libpq` and `psycopg2` versions in `group_vars/all` were being overridden by hardcoded values in `bee-pollination.base/meta/main.yml`. As a result, the App servers are/have been using `psycopg2==2.6.*` and `libpq-dev=9.3.*`. According to https://github.com/project-icp/bee-pollinator-app/commit/5fe0b9500c7b9507d40c042bc325c831f5df4b95  that's the version that we _should_ be at, so I chose to use those versions instead of the newer ones in `group_vars/all`

# Deployment checklist
- [x] deploy using Jenkins CI pipeline
- [x] run migrations
- [x] Add Admin user
        - [ ] Save credentials in LastPass
- [x] Verify new user sign up and email functionality
- [x] Verify modeling endpoint functionality
- [x] Verify project creation/sharing functionality

# Testing
- I deployed this project to Staging using Jenkins. See http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-develop/68/console
- See https://staging.app.pollinationmapper.org
- Verify ELK stack is functioning at ~https://monitoring.staging.app.pollinationmapper.org:5601~ http://monitoring.staging.app.pollinationmapper.org:5601

Issue #288, opened 2018-10-04
Add Endpoint for Forage Rasters


rajadain:
- [x] Add sample PA rasters from `smb://fileshare.internal.azavea.com/projects/PennStateUniversity_USDABeekeepers-FFAR/data/USDA_AFRI_azavea` to the VM or S3 (wherever the previous CDL raster is stored)
- [x] Decide how the Beekeepers URLs and Views will work
  - Consider adding `views.py`, `urls.py`, `tasks.py` under the `beekeepers` app added in #289 
- [x] Add an endpoint (similar to [`start_crop_yield`](https://github.com/project-icp/bee-pollinator-app/blob/1cf3c26d54d76ee596c33e370b48bc3722f57378/src/icp/apps/modeling/views.py#L173-L180)) that takes a point and buffer range, and queries the sample PA forage rasters in all three seasons, and returns the values in each. It may be possible to avoid using Celery, if the raster reads are quick.

  - This should return a JSON result like this:

    ```json
    {
      "forage_summer": 23,
      "forage_spring": 14,
      "forage_fall": 18
    }
    ```

PR #305, opened 2018-10-23
Beekeeper raster endpoint

fungjj92:
## Overview

The beekeeper app will need to query rasters in s3 to get data at a point. The rasters are pre-processed by the client for single-value lookups and the data sets may be updated by the client directly on s3. This PR setups up beekeeper's first API endpoint with this functionality.

Connects #288 

### Demo
screen shot 2018-10-23 at 3 01 15 pm

### Notes

Wow, AWS permissions were a serious doozy and delayed this PR significantly.
Rasterio making requests to s3 requires AWS IAM credentials. While boto3 under the hood of Rasterio automatically picks up various env variables pointing to or specifying AWS cred info, several manual steps by Ansible were necessary to make the AWS creds file readable from within Django by boto3.
- Correct `read` permissions on `~/.aws/` and its contents
- Mounting `~/.aws` to the correct user `$HOME` dir in the `app` VM (the user is `icp` and said user is created by Ansible so isn't avail until provision is run)
- Ensuring the correct permissions on `~/.aws` as the `icp` user within the VM
- Setting Ansible env variables (particularly `AWS_PROFILE` because boto3 picks that up automatically to read `~/.aws/credentials`)
- Making available an HTTPS cert required by rasterio to query s3 (`/etc/ssl/certs/ca-certificates.crt`)
- Working around this rasterio [clobbered creds issue ](https://github.com/mapbox/rasterio/issues/1507)(fixed as of 10/24/18)

Known follow-up work (different cards):
- Have the endpoint take geom, forage-range, and filenames data
- Getting the homepage and the beekeeper app urls on the same url scheme (currently `?beekeepers` and `/beekeepers`). The [ADR discussed ](https://github.com/project-icp/bee-pollinator-app/blob/develop/doc/phase-2/000_project_architecture_adr.md#consequences)something like the latter.

### Testing Instructions
- Get AWS IAM credentials for Bees and add them to an `icp_bees` profile in your `~/.aws/credentials` file
- Follow the Beekeepers section of the README (if anything is unclear or doesn't work, good time for feedback). If you haven't rebuilt your VM since before #294 I suggest destroying your VM first.
- Once you do the above, hit http://localhost:8000/beekeepers/fetch/ and click "Post". 
- You should get back data
- Feel free to play with the `geom` or `forage_range` in `beekeepers/tasks.py` to test different data

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,9 @@ Beekeepers is a separate front-end app that takes advantage of the API infrastru
 
 The app can be viewed at [http://localhost:8000/?beekeepers](http://localhost:8000/?beekeepers).
 
+## AWS Credentials
+Raster data used by Beekeeper is stored and accessed from a private s3 bucket. To authenticate your requests, get your own IAM credentials and add them under an `icp_bees` profile to `~/.aws/credentials`. Ensure your `~/.aws/credentials` file has at least read permissions, else run `chmod -R 644 ~/.aws`. Provision your VM to mount the AWS credentials with Ansible.
+
 To build the Beekeepers App, run:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ Beekeepers is a separate front-end app that takes advantage of the API infrastru
 The app can be viewed at [http://localhost:8000/?beekeepers](http://localhost:8000/?beekeepers).
 
 ## AWS Credentials
-Raster data used by Beekeeper is stored and accessed from a private s3 bucket. To authenticate your requests, get your own IAM credentials and add them under an `icp_bees` profile to `~/.aws/credentials`. Ensure your `~/.aws/credentials` file has at least read permissions, else run `chmod -R 644 ~/.aws`. Provision your VM to mount the AWS credentials with Ansible.
+Raster data used by Beekeeper is stored and accessed from a private s3 bucket. To authenticate your requests, get your own IAM credentials and add them under an `icp-bees` profile to `~/.aws/credentials`. Ensure your `~/.aws/credentials` file has at least read permissions, else run `chmod -R 644 ~/.aws`. Provision your VM to mount the AWS credentials with Ansible.
 
 To build the Beekeepers App, run:
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -112,6 +112,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     app.vm.hostname = "app"
     app.vm.network "private_network", ip: ENV.fetch("ICP_APP_IP", "33.33.34.10")
 
+    app.vm.synced_folder "~/.aws", "/home/vagrant/.aws", type: "rsync"
+    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync", owner: "icp", group: "icp", :mount_options =>[dmode=644,fmode=644]
+
     if Vagrant::Util::Platform.windows? || Vagrant::Util::Platform.cygwin?
       app.vm.synced_folder "src/icp", "/opt/app/", type: "rsync", rsync__exclude: ["node_modules/", "apps/"]
       app.vm.synced_folder "src/icp/apps", "/opt/app/apps"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -113,7 +113,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     app.vm.network "private_network", ip: ENV.fetch("ICP_APP_IP", "33.33.34.10")
 
     app.vm.synced_folder "~/.aws", "/home/vagrant/.aws", type: "rsync"
-    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync", owner: "icp", group: "icp", :mount_options =>[dmode=644,fmode=644]
+    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync"
 
     if Vagrant::Util::Platform.windows? || Vagrant::Util::Platform.cygwin?
       app.vm.synced_folder "src/icp", "/opt/app/", type: "rsync", rsync__exclude: ["node_modules/", "apps/"]
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -23,6 +23,7 @@ celery_processes_per_worker: 1
 
 docker_options: "-H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock --storage-driver=aufs"
 
-aws_profile: "icp-stg"
+aws_profile: "icp_bees"
+aws_beekeepers_data_bucket: "beekeepers-staging-data-us-east-1"
 
 stack_type: "Development"
--- a/deployment/ansible/group_vars/development
+++ b/deployment/ansible/group_vars/development
@@ -23,7 +23,7 @@ celery_processes_per_worker: 1
 
 docker_options: "-H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock --storage-driver=aufs"
 
-aws_profile: "icp_bees"
+aws_profile: "icp-bees"
 aws_beekeepers_data_bucket: "beekeepers-staging-data-us-east-1"
 
 stack_type: "Development"
--- a/deployment/ansible/roles/bee-pollinator.base/defaults/main.yml
+++ b/deployment/ansible/roles/bee-pollinator.base/defaults/main.yml
@@ -14,3 +14,5 @@ envdir_config:
   ROLLBAR_SERVER_SIDE_ACCESS_TOKEN: "Unknown"
   AWS_PROFILE: "{{ aws_profile }}"
   AWS_BEEKEEPERS_DATA_BUCKET: "{{ aws_beekeepers_data_bucket }}"
+  CURL_CA_BUNDLE: "{{ ubuntu_ssl_certs }}"
+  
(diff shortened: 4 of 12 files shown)


fungjj92:
Tagging @tnation14 as well since there are some Ansible changes

README.md
```diff
@@ -274,6 +274,9 @@ Beekeepers is a separate front-end app that takes advantage of the API infrastru
 
 The app can be viewed at [http://localhost:8000/?beekeepers](http://localhost:8000/?beekeepers).
 
+## AWS Credentials
+Raster data used by Beekeeper is stored and accessed from a private s3 bucket. To authenticate your requests, get your own IAM credentials and add them under an `icp_bees` profile to `~/.aws/credentials`. Ensure your `~/.aws/credentials` file has at least read permissions, else run `chmod -R 644 ~/.aws`. Provision your VM to mount the AWS credentials with Ansible.
```
tnation14: If we modify the shared folder group/owner with Vagrant as explained below, we can skip the `chmod` step. I was able to get things working with `600` permissions on `~/.aws` on the host machine.

Vagrantfile
```diff
@@ -112,6 +112,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     app.vm.hostname = "app"
     app.vm.network "private_network", ip: ENV.fetch("ICP_APP_IP", "33.33.34.10")
 
+    app.vm.synced_folder "~/.aws", "/home/vagrant/.aws", type: "rsync"
+    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync"
```
tnation14: I was able to use the raster endpoint after setting ownership on `/var/lib/icp/.aws` with the following diff:

```diff
diff --git a/Vagrantfile b/Vagrantfile
index c9ce1238..728339b3 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -113,7 +113,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     app.vm.network "private_network", ip: ENV.fetch("ICP_APP_IP", "33.33.34.10")
 
     app.vm.synced_folder "~/.aws", "/home/vagrant/.aws", type: "rsync"
-    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync"
+    app.vm.synced_folder "~/.aws", "/var/lib/icp/.aws", type: "rsync", owner: "icp", group: "icp"
 
     if Vagrant::Util::Platform.windows? || Vagrant::Util::Platform.cygwin?
       app.vm.synced_folder "src/icp", "/opt/app/", type: "rsync", rsync__exclude: ["node_modules/", "apps/"]
diff --git a/deployment/ansible/roles/bee-pollinator.beekeepers/tasks/main.yml b/deployment/ansible/roles/bee-pollinator.beekeepers/tasks/main.yml
index ed4643fc..b4b0efab 100644
--- a/deployment/ansible/roles/bee-pollinator.beekeepers/tasks/main.yml
+++ b/deployment/ansible/roles/bee-pollinator.beekeepers/tasks/main.yml
@@ -36,10 +36,4 @@
 - name: Link Static Directory
   file: path="{{ beekeepers_static_root }}"
         src="{{ beekeepers_home }}/dist"
-        state=link
-
-- name: Change AWS credentials access to the icp user
-  file: path="/var/lib/icp/.aws"
-        group="icp"
-        owner="icp"
-  when: "['development', 'test'] | some_are_in(group_names)"
+        state=link
```

I think it'd be cleaner to let Vagrant set the file's owners at boot rather than "fixing" ownership with Ansible after the VM is spun up.

See https://www.vagrantup.com/docs/synced-folders/basic_usage.html#modifying-the-owner-group for more info on configuring permissions for Vagrant Shared Folders.

deployment/ansible/roles/bee-pollinator.beekeepers/tasks/main.yml
```diff
@@ -37,3 +37,9 @@
   file: path="{{ beekeepers_static_root }}"
         src="{{ beekeepers_home }}/dist"
         state=link
+
+- name: Change AWS credentials access to the icp user
```
tnation14: If file ownership is set with Vagrant, this task can be removed.

deployment/ansible/roles/bee-pollinator.base/defaults/main.yml
```diff
@@ -12,3 +12,7 @@ envdir_config:
   ICP_STACK_COLOR: "{{ stack_color }}"
   ICP_STACK_TYPE: "{{ stack_type }}"
   ROLLBAR_SERVER_SIDE_ACCESS_TOKEN: "Unknown"
+  AWS_PROFILE: "{{ aws_profile }}"
+  AWS_BEEKEEPERS_DATA_BUCKET: "{{ aws_beekeepers_data_bucket }}"
+  CURL_CA_BUNDLE: "{{ ubuntu_ssl_certs }}"
```
tnation14: 🔥

Vagrantfile
fungjj92: This was my original solution actually! However, I moved ownership assignment to the Ansible task because the `icp` VM user is created by Ansible during provision. If you are building the `app` VM from scratch, you will not have the `icp` user yet and the added Vagrantfile line will fail.

tnation14: Ah, makes sense; I missed that case because didn't build `app` from scratch.

deployment/ansible/group_vars/development
```diff
@@ -23,6 +23,7 @@ celery_processes_per_worker: 1
 
 docker_options: "-H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock --storage-driver=aufs"
 
-aws_profile: "icp-stg"
+aws_profile: "icp_bees"
```
rajadain: Minor thing, but can this be `icp-bees` instead of `icp_bees`? All the other projects use hyphens, not underscores.

fungjj92: I swear I got `icp_bees` from somewhere. I can't remember where that somewhere is now.

src/icp/apps/beekeepers/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+        'forage_fall_5k.tif'
+    ]
+}
+
+
+def sample_at_point(geom, raster_path):
+    """
+    Return the cell value for a raster at a provided lat lon coordinate
+    Args:
+        geom({x: float, y: float}): A EPSG:4326 point in the same SRS as the
```
rajadain: Prefer using [standard docstrings](https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html#auto-documented-class-method-1):

```python
"""
Return the cell value for a raster at a provided lat lon coordinate

:param geom: A EPSG:4326 point in the same SRS as the target raster
             defining the point to extract the cell value.
:param raster_path: A file path to a geographic raster containing the
                    value to extract. String.
:type geom: Dict with x and y values, like {x: float, y: float}.
:type raster_path: String
:return: The cell value at the point geom. Same data type as raster.
"""
```

rajadain: Also consider switching `x` and `y` to `lat` and `lon` (or `lon` and `lat`, however it fits).

src/icp/apps/beekeepers/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+            target raster defining the point to extract the cell value.
+        raster_path (string): A file path to a geographic raster
+            containing the value to extract.
+    Returns:
+        The cell value, in the data type of the input raster, at the point geom
+    """
+    with rasterio.open(raster_path, mode='r') as src:
```
rajadain: Consider adding a `# TODO` note here to catch errors with opening rasters in the future.

src/icp/apps/beekeepers/tasks.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Sample the raster at the given coordinates
+        value_gen = src.sample([(x, y)], indexes=[1])
+        value = value_gen.next().item(0)
+        return value
+
+
+@decorators.api_view(['POST'])
+@decorators.permission_classes((AllowAny, ))
+def fetch_data(request):
```
rajadain: Consider moving this view into `views.py`.

src/icp/apps/beekeepers/tasks.py
mmcfarland: Actually, `A EPSG:4326 point in the same SRS as the target raster` is a bit contradictory.  If it's in the target SRS, it likely won't be 4326.  Is this actually a `Shapely geometry` type?  If it's not incoming as geographic coordinates, `x` & `y` are likely more accurate.  

Also, since we know the target CRS of the source rasters, we can probably embed that in the documentation directly, instead of referring to it generally.

mmcfarland: Looking at the code, it seems we _are_ reprojecting it to the crs of the raster and assuming it's in 4326.  In that case, I'd just drop the "in the same SRS as the target raster" segment to keep it accurate.

fungjj92:
Thanks for the reviews y'all, I've pushed up changes.

.gitignore
```diff
@@ -100,3 +100,6 @@ _SUCCESS
 
 # Webpack Stats used to help Django Hot Reload
 .webpack-stats.json
+
+#vscode
+.vscode/
```
rajadain: For developer-specific ignores, prefer using a [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) file.

src/icp/apps/beekeepers/views.py
```diff
@@ -0,0 +1,53 @@
...
+@decorators.permission_classes((AllowAny, ))
+def fetch_data(request):
+    """
+    Return the cell values from rasters on s3
+
+    TODO: Receives validated geom, forage range, and filename(s) from the
+    request
+    """
+    geom = {
+        'x': 39.9473832311,
```
rajadain: Should these be also changed to `lat` and `lon`?

fungjj92: doh

src/icp/apps/beekeepers/tasks.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import division
+
+import os
```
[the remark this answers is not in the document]
fungjj92: oooooo 👍

fungjj92:
CI succeeded! Merging. Thanks for review :)

Issue #296, opened 2018-10-19
Setup Redux


rajadain:
Redux is already included in the sample app [here](https://github.com/project-icp/bee-pollinator-app/blob/develop/src/icp/apps/beekeepers/js/src/reducers.js).

- [x] Add an initial state as sketched out in https://github.com/project-icp/bee-pollinator-app/issues/292#issuecomment-431479309
- [x] Setup Redux Persist to save the Redux state to localStorage

Issue #292, opened 2018-10-18
Beekeepers UI Handoff and Planning


rajadain:

mmcfarland:
If `forageRange` was changed from `3` to `5`, would a new entry be added to `scores` after calculation

```json
"forage_spring_5k": {
    "data": 75,
    "error": null
}
```

Such that on subsequent flips of that dropdown, we check and see we already have a value?

rajadain:
Good point, we got confused and thought that the `3k` and `5k` applied only to the forage rasters. In fact, the seasons only apply to forage rasters, but the forage range applies to all rasters. In this case, we may use that as a the first key under `scores`, like this:

```json
{
    "sortBy": "",
    "forageRange": "",
    "apiaries": [
        {
            "name": "",
            "location": {
                "lat": 1.0,
                "lon": 1.0
            },
            "scores": {
                "3k": {
                    "hive_density": {
                        "data": 26,
                        "error": null
                    },
                    "forage_fall": {
                        "data": 45,
                        "error": null
                    },
                    "forage_spring": {
                        "data": 61,
                        "error": null
                    },
                },
                "5k": {
                    "hive_density": {
                        "data": 30,
                        "error": null
                    },
                    "forage_fall": {
                        "data": 49,
                        "error": null
                    },
                    "forage_spring": {
                        "data": 65,
                        "error": null
                    },
                }
            },
            "fetching": false,
            "starred": false,
            "surveyed": false,
        }
    ]
}
```

The API will be queried with a given forage range and those results will be saved the first time. In the future, if the values are already there, they will be shown instantly.

Example:

- Forage Range is set to "3km"
- User makes a point **A** on the map
- Scores for **A** are fetched with a forage range for "3km"
- User switches Forage Range to "5km"
- Scores for **A** are fetched with a forage range for "5km"
- User adds a new point **B** on the map
- Scores for **B** are fetched with a forage range of "5km"
- User switches Forage Range to "3km"
- Scores for **A** with a forage range of "3km" are shown instantly,
  Scores for **B** are fetched with a forage range of "3km"

mmcfarland:
Sounds good.  Additionally, we'll be able to play with the idea of fetching all 3k/5k values at one go depending on performance.

rajadain:
For prop type validation, it may be better for us to go with this structure for `scores`:

```json
"scores": {
    "3k": [
        {
            "name": "hive_density",
            "data": 26,
            "error": null
        },
        {
[...]

PR #310, opened 2018-10-26
Initialize Redux with Persist

rajadain:
## Overview

Adds basic Redux setup using the [sample state](https://github.com/project-icp/bee-pollinator-app/issues/292#issuecomment-431479309) and Redux Persist. The setup for Redux Persist was taken mostly from [UWR](https://github.com/azavea/pwd-urban-water-revealed/blob/1.0.0/src/app/js/src/store.js), with some changes recommended in the [v5 Migration Guide](https://github.com/rt2zz/redux-persist/blob/208c6b112ead87b3701dfacaae2cdbe78377775a/docs/MigrationGuide-v5.md), particularly around `autoRehydrate`.

Connects #296 

### Demo

image

## Testing Instructions

* Check out this branch
* Run `./scripts/beekeepers.sh install` to update yarn libraries
* Run `./scripts/beekeepers.sh start` to run the app and go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) to see it
* Inspect the console. Ensure you see `persist/` actions.
* Inspect the localStorage. Ensure you see the initial state saved there.


src/icp/apps/beekeepers/package.json
```diff
@@ -35,18 +35,19 @@
     "redux": "~3.7.2",
     "redux-act": "~1.7.4",
     "redux-logger": "~3.0.6",
+    "redux-persist": "^5.10.0",
```
kellyi: I think the initial `^` should be a `~`

src/icp/apps/beekeepers/js/src/store.js
```diff
@@ -11,4 +15,12 @@ if (process.env.NODE_ENV === 'development') {
...
 const createStoreWithMiddleware = applyMiddleware(...middlewares)(createStore);
 
-export { createStoreWithMiddleware as default };
+const persistReducerConfig = {
+    key: 'beekeepers-app',
+    storage,
+};
+const persistReducers = persistCombineReducers(persistReducerConfig, reducers);
+const store = createStoreWithMiddleware(persistReducers);
+const persistor = persistStore(store);
```
kellyi: Apologies for floating by with comments but just wanted to note that this will persist the entirety of the Redux store to localStorage. Could be that that's what's desirable here but alternatively you can persist only certain reducers:

https://github.com/rt2zz/redux-persist#nested-persists

rajadain: I appreciate it! Was hoping someone with more experience would take a look. I'll look into that.

rajadain: Rather than use per-reducer persist configuration, I added an option that will only include the specified reducers in 1aa7713.

rajadain:
This is ready for a look now.

src/icp/apps/beekeepers/js/src/main.jsx
```diff
@@ -3,19 +3,21 @@ import { render } from 'react-dom';
...
-import reducers from './reducers';
+import { store, persistor } from './store';
 
-const store = createStoreWithMiddleware(reducers);
 render(
     <Provider store={store}>
-        <BrowserRouter>
-            <Route exact path="/" component={App} />
-        </BrowserRouter>
+        {/* TODO Add Loading State */}
```
fungjj92: Is this a copy over from UWR? If it's something we want, we should card it. I vote Xing or punting over to the enhancement stage.

fungjj92:
Looks good! A question. There's no duration on how long we persist the app data in localStorage. Is that intentional?

src/icp/apps/beekeepers/js/src/store.js
kellyi: Makes sense. Do we know what parts of the state it would be desirable to persist in localStorage? And conversely what pieces will be saved in the database via Django?

rajadain: I think the idea is that if the user does not have an account, we persist to localStorage. If they do have an account, we use the database. The exact difference will hopefully be revealed as we build up, but this seems enough for a base setup.

src/icp/apps/beekeepers/js/src/main.jsx
rajadain: I'll make a card and add a link to it in the comment.

rajadain:
> There's no duration on how long we persist the app data in localStorage. Is that intentional?

Yes. The idea is that if the user does not have an account, we store this in localStorage. Unlike UWR, which is most often a one-time activity, this app is something that users may return to after a long time (such as seasons). The exact nature of when and what to store (and when not to) should become clearer as we develop the user account part of the app, and this bootstrapping allows us to move in any direction when we need to.

src/icp/apps/beekeepers/js/src/store.js
kellyi: > The exact difference will hopefully be revealed as we build up, but this seems enough for a base setup.

It seems like there's a medium-to-strong case for not adding redux-persist now but instead waiting to add it later until it's been determined what needs to be stored and what doesn't.

For example, in the abstract I don't think it makes sense to persist `fetching` and `error` states, but the way this is set up will make that happen -- and might compel writing some additional component lifecycle code to deal with that.

Additionally, I think that adding Redux persist now will create the possible necessity of implementing some kind of versioning scheme for the store if the shape of the main reducer changes. (In itself this isn't too problematic since I imagine there may only be once production release, but it would impact staging.)

fungjj92: I see Kelly's points and had the work here not already been done, pushing back redux-persist would be fine. But i don't think it's a bad thing that it's here. I'm not so concerned about what pieces of data to save because that can be cherry-picked later, as we add shape and real data to the state. Staging conflict of state can be resolved by clearing the localStorage. We added redux-persist earlyish to UWR and it was helpful & not a nuisance.

rajadain: After conversing in real person, I've added a separate, empty reducer called `saved` to hold items that we explicitly want to persist. This allows us to keep the redux-persist setup without impeding the evolution of `main` state. Work done in 0b79ef4.

src/icp/apps/beekeepers/js/src/reducers.js
```diff
@@ -60,6 +60,10 @@ const mainReducer = createReducer({
         (state, payload) => update(state, { forageRange: { $set: payload } }),
 }, initialState);
 
+// Placeholder reducer for parts of state that will be persisted to localStorage
+const savedReducer = createReducer({}, {});
+
 export default {
```
fungjj92: I think we're looking for the `combineReducers` function here? https://redux.js.org/api/combinereducers

rajadain: No. We're using [`persistCombineReducers`](https://github.com/project-icp/bee-pollinator-app/pull/310/files#diff-e8187819ec48e38e3cdde21227a54e87R23) according to the [redux-persist migration guide](https://github.com/rt2zz/redux-persist/blob/208c6b112ead87b3701dfacaae2cdbe78377775a/docs/MigrationGuide-v5.md), so this can just be a simple object.

rajadain:
Thanks for reviewing! Squashed the fixups, merging this now.

Issue #297, opened 2018-10-19
Create Initial Components


rajadain:
- [x] Make a component for the Sidebar
- [x] Make a component for the Splash Page
  - This will just be static content as shown here:

      ![image](https://user-images.githubusercontent.com/1430060/47241463-be4e1980-d3b9-11e8-9c6c-c25b207cf7ba.png)

  - The splash page will be shown when there are no apiaries, i.e. the `apiaries` list in Redux is empty

PR #312, opened 2018-10-29
Add Splash, Sidebar component

rajadain:
## Overview

Adds a Splash component with basic styling. Extracts Map component into its own placeholder. Adjusts Prototype component to include only the sidebar contents for now. Adjusts App component accordingly.

Connects #297 

### Demo

image

Tagging @jfrankl 	for visual and style review.

#### BEM

Am I following BEM correctly for naming CSS classes?

```html
<div className="splash__info_box">
    <div className="splash__info_box__title">
        Get Started
    </div>
    <div className="splash__info_box__body">
        Click on the map to see how it stacks up
    </div>
</div>
```

It feels a little redundant having to specify `splash__info_box` multiple times.

#### `stylelint`

I liked that `stylelint` will auto-format the SCSS files on save, including the order of CSS rules, as long as `beekeepers start` is running.

#### Mouse Cursor Icon

I did not see an icon or image in the prototype for the mouse cursor click icon shown in the mockup:

image

Any recommendations on where to find it?

#### Style Variable Use

I got the page to be as close to the prototype as I could, but it's not quite identical yet in terms of text size, and it was taking me too long to tune the details, so I've left it as a design task for later. One thing that I did do, however, is use the existing type size variables wherever I could:

```sass
&__header {
    margin-top: 1rem;
    margin-bottom: $font-size-xl;
    font-weight: 900;
    font-size: $font-size-xxl;
    line-height: $font-size-3xl;
}
```

even in places that aren't strictly type related, such as `margin-bottom` above. This was done in an effort to maintain the established rhythm, but if this is seen as bad practice I can change those out to be literal values.

## Testing Instructions

* Check out this branch and `./scripts/beekeepers.sh start`
* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and ensure you see the splash screen


src/icp/apps/beekeepers/js/src/components/Splash.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+export default function () {
+    return (
+        <div className="sidebar splash">
+            <h2 className="splash__header">
+                Locate your next bee hive anywhere in the world
```
fungjj92: 2 details:
The top level class here can be "splash" to decouple the component from the `sidebar` and avoid potential css conflicts
I'm not sure, however I think bee hive is one word "beehive"

src/icp/apps/beekeepers/js/src/components/Sidebar.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { arrayOf, object } from 'prop-types';
+import { connect } from 'react-redux';
+
+import Splash from './Splash';
+import Prototype from './Prototype';
+
+function mapStateToProps(state) {
```
fungjj92: For organization, can we move `mapStateToProps` just above `export default connect()();` ? I think it is convention that the component  (`Sidebar`) is at the top.

src/icp/apps/beekeepers/js/src/components/Splash.jsx
rajadain: I am reusing some `sidebar` CSS though. Can duplicate those rules in `splash` if that's preferable.

fungjj92: Ah. That's fine then.

Issue #299, opened 2018-10-19
Add Basemap


rajadain:
In the wireframes, the Mapbox Satellite Streets basemap is used:

![image](https://user-images.githubusercontent.com/1430060/47241463-be4e1980-d3b9-11e8-9c6c-c25b207cf7ba.png)

Mapbox basemaps typically require tokens. Should we get the token from the client? Should we use a different basemap?

fungjj92:
Requires #300

mmcfarland:
We had decided to use Esri world imagery (same as Pollinator) but try adding the Carto label layer as a composite as well.

Issue #298, opened 2018-10-19
Create Map Component


rajadain:
- [x] Evaluate if React Leaflet suffices for the needs of this project
- [x] Make a Map component using React Leaflet or whatever else as decided above

PR #315, opened 2018-10-31
Add map to the map component

fungjj92:
## Overview

Add a map to the existing map component added in #312. I added a few map properties to the initial state, subject to change. It hasn't be determined what the map interactions will be yet.

Use OSM tiles for now, basemap TBD in #299 

Connects #298 

### Demo

screen shot 2018-10-31 at 9 32 01 am

## Testing Instructions

`./scripts/beekeepers.sh start` and the map should appear.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/reducers.js
+++ b/src/icp/apps/beekeepers/js/src/reducers.js
@@ -1,19 +1,12 @@
 import { createReducer } from 'redux-act';
 import update from 'immutability-helper';
 
-import {
-    SORT_CREATED,
-    FORAGE_RANGE_3K,
-    MAP_CENTER,
-    MAP_ZOOM,
-} from './constants';
+import { SORT_CREATED, FORAGE_RANGE_3K } from './constants';
 import { setSort, setForageRange } from './actions';
 
 const initialState = {
     sortBy: SORT_CREATED,
     forageRange: FORAGE_RANGE_3K,
-    mapCenter: MAP_CENTER,
-    zoom: MAP_ZOOM,
     apiaries: [
         /* Sample Apiary Structure
         {
--- a/src/icp/apps/beekeepers/js/src/reducers.js
+++ b/src/icp/apps/beekeepers/js/src/reducers.js
@@ -1,12 +1,19 @@
 import { createReducer } from 'redux-act';
 import update from 'immutability-helper';
 
-import { SORT_CREATED, FORAGE_RANGE_3K } from './constants';
+import {
+    SORT_CREATED,
+    FORAGE_RANGE_3K,
+    MAP_CENTER,
+    MAP_ZOOM,
+} from './constants';
 import { setSort, setForageRange } from './actions';
 
 const initialState = {
     sortBy: SORT_CREATED,
     forageRange: FORAGE_RANGE_3K,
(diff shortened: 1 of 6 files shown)


src/icp/apps/beekeepers/js/src/reducers.js
```diff
@@ -1,12 +1,19 @@
...
+    MAP_CENTER,
+    MAP_ZOOM,
+} from './constants';
 import { setSort, setForageRange } from './actions';
 
 const initialState = {
     sortBy: SORT_CREATED,
     forageRange: FORAGE_RANGE_3K,
+    mapCenter: MAP_CENTER,
+    zoom: MAP_ZOOM,
```
rajadain: Do we need to store the map center and zoom in redux? I think this could just be internal state managed by the component itself.

src/icp/apps/beekeepers/templates/beekeepers/beekeepers.html
```diff
@@ -6,6 +6,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta content="IE=edge" http-equiv="X-UA-Compatible">
     <title>Beekeepers App</title>
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/normalize/7.0.0/normalize.min.css">
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.1/leaflet.css">
```
rajadain: Will we have to keep in mind to update this if we ever update Leaflet? Is this CSS not available in the Leaflet package?

src/icp/apps/beekeepers/js/src/reducers.js
fungjj92: Good question. I guess once the user has data aka apiaries, we'd initialized the map to the extent of the points. If there's no apiaries we can just start at this dummy point. That can be managed by the map component.

src/icp/apps/beekeepers/templates/beekeepers/beekeepers.html
fungjj92: Yes to the former. This was part of the react-leaflet documentation. It seems to be how react-leaflet works, which is odd because I'm not sure what the case is that you wouldn't want these CSS files. For example, they provide the critical function to arrange and align the map tiles.

rajadain: Oh we do want those CSS files, the difference is between getting them from our build vs the CDN.

For example, we've done this in the past: https://github.com/azavea/pgw-community-mapping/blob/ab48f0f1a4c0e2f45fbffeb02909f612acbb4d4e/src/app/sass/main.scss#L1, to get it from the installed package.

src/icp/apps/beekeepers/js/src/reducers.js
rajadain: I agree! I think we should use the default values from `constants` directly in the component and just omit them from the redux state.

src/icp/apps/beekeepers/templates/beekeepers/beekeepers.html
fungjj92: The CSS files don't come as part of the react-leaflet package

Issue #302, opened 2018-10-19
Make Apiary Card Component


rajadain:
Make a component for an Apiary card as shown:

![image](https://user-images.githubusercontent.com/1430060/47241970-74fec980-d3bb-11e8-9c1f-814213627ccc.png)

Apiaries should be listed in the order in which they were added, which should be the same order in the Redux state.

Make the delete button remove the apiary from the Redux state.

Use placeholders for scores.

PR #317, opened 2018-10-31
Make Apiary Card Component

rajadain:
## Overview

Breaks down the Apiary Card into a number of components, and populates them via state. Also added some dummy state to populate the cards.

Connects #302 

### Demo

image

## Testing Instructions

* Check out this branch and `beekeepers build`
* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and see the cards


src/icp/apps/beekeepers/js/src/components/CardButton.jsx
```diff
@@ -0,0 +1,28 @@
...
+};
+
+CardButton.propTypes = {
+    icon: string.isRequired,
+    filled: bool.isRequired,
+    onClick: func,
+};
+
+CardButton.defaultProps = {
+    onClick: () => {},
```
fungjj92: We don't need the onClick bits now in this component

src/icp/apps/beekeepers/js/src/propTypes.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const Scores = shape(RASTERS.reduce(
+    (map, raster) => Object.assign(map, { [raster]: Score.isRequired }),
+    {},
+));
+
+export const Apiary = shape({
+    name: string.isRequired,
+    marker: string.isRequired,
+    location: LatLon.isRequired,
```
fungjj92: hm during development, requiring all of these Apiary fields will be infeasible. Some can be required, though, like `location`.

src/icp/apps/beekeepers/js/src/components/ScoreLabelTriplet.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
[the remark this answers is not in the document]
rajadain: Haha that's quite the discovery. Will do.

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
```diff
@@ -0,0 +1,93 @@
...
+        marker,
+        name,
+        selected,
+        starred,
+        surveyed,
+        scores: { threeKm },
+    } = apiary;
+
+    const markerMod = (() => {
+        if (surveyed) {
```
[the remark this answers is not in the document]
rajadain: I was going off the CSS classes in the prototype for now:

https://github.com/project-icp/bee-pollinator-app/blob/afba6f5c24678ec8074981cd2eac45b84dea3003/prototype/beekeepers/src/sass/06_components/_marker.scss#L11-L21

but didn't spend too much time thinking about it because it is in flux.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -3,5 +3,22 @@ export const FORAGE_RANGE_5K = '5k';
 
 export const SORT_CREATED = 'SORT_CREATED';
 
+export const RASTER_HIVE_DENSITY = 'hive_density';
```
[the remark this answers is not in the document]
rajadain: I see your point. We may save some space by defining it as an object. Will switch to that.

src/icp/apps/beekeepers/js/src/propTypes.js
rajadain: We can use a placeholder address for the `name`. The `marker` is just a monotonically incrementing letter (A, B, C, ...), and we can use a placeholder `A` there as well. Why is it infeasible?

fungjj92: Hm I assumed types here operate like in typescript. In TS, type properties labelled as required if not fulfilled by an object defined as that type will cause app failure. Can an obj of type `Apiary` omit some of the listed properties and be valid?

rajadain: No it cannot, and neither should it, since we need to show the name and the marker for each card. But fulfilling them should be quite feasible.

fungjj92: A clarification. I wasn't referring to the fields above this comment in particular, rather `Apiary` as a whole. I could have placed my comment elsewhere for better clarity.

In general where I'm coming from is that strict and overly-detailed types early during development can be an unhelpful amount of structure to fulfill.

src/icp/apps/beekeepers/js/src/components/CardButton.jsx
rajadain: This is so we can pass in a handler for the delete apiary button.

src/icp/apps/beekeepers/js/src/propTypes.js
rajadain: Following an in-person conversation, I'm deciding to keep these since otherwise we'd have to think of empty states in the UI (when these props are not supplied).

rajadain:
The fixups apply the suggestions to the codebase. This is ready for another look!

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { arrayOf, string } from 'prop-types';
+
+import { Score } from '../propTypes';
+import { toDashedString, toSpacedString } from '../utils';
+
+const ScoreLabelTriplet = ({ raster, scores }) => (
```
fungjj92: Ctrl+f for `ScoreLabelTriplet` --> `ScoresLabel`

src/icp/apps/beekeepers/js/src/components/Sidebar.jsx
```diff
@@ -1,19 +1,47 @@
...
+                <div className="controls__select">
+                    <select>
+                        <option value="one">one</option>
+                        <option value="two">two</option>
+                        <option value="three">three</option>
+                    </select>
+                </div>
+            </div>
+            <ul className="card-container">
+                {apiaries.map(apiary => <ApiaryCard key={apiary.marker} apiary={apiary} />)}
```
fungjj92: Can we move this mapping operation to a variable before the `return`?

rajadain:
Thanks for reviewing! Squashing the fixups now, will wait for that to pass CI before merging.

rajadain:
Finally green! Merging this now. 🥗

Issue #300, opened 2018-10-19
Add Map Interactions


rajadain:
- [x] Show markers on the map for apiaries in the Redux State
- [x] Make clicking on the map add an apiary

rajadain:
Consider using the revised apiary state structure from #317:

```js
const initialState = {
    sortBy: SORT_CREATED,
    forageRange: FORAGE_RANGE_3K,
    apiaries: [
        // TODO Remove dummy data
        {
            name: '423 Waltz Road',
            marker: 'A',
            location: {
                lat: 1.0,
                lon: 1.0,
            },
            scores: {
                threeKm: {
                    [RASTER_HIVE_DENSITY]: { data: 26, error: null },
                    [RASTER_HABITAT]: { data: 13, error: null },
                    [RASTER_PESTICIDE]: { data: 20, error: null },
                    [RASTER_FORAGE_SPRING]: { data: 61, error: null },
                    [RASTER_FORAGE_SUMMER]: { data: 54, error: null },
                    [RASTER_FORAGE_FALL]: { data: 45, error: null },
                    [RASTER_OVERALL]: { data: 45, error: null },
                },
                fiveKm: {
                    [RASTER_HIVE_DENSITY]: { data: 26, error: null },
                    [RASTER_HABITAT]: { data: 13, error: null },
                    [RASTER_PESTICIDE]: { data: 20, error: null },
                    [RASTER_FORAGE_SPRING]: { data: 61, error: null },
                    [RASTER_FORAGE_SUMMER]: { data: 54, error: null },
                    [RASTER_FORAGE_FALL]: { data: 45, error: null },
                    [RASTER_OVERALL]: { data: 45, error: null },
                },
            },
            fetching: false,
            selected: false,
            starred: false,
            surveyed: false,
        },
    ],
};
```

PR #319, opened 2018-11-02
Map click adds marker to map and updates state

fungjj92:
## Overview

The app user's map usage will be pretty simple -- if the user clicks the map, an apiary marker should be added. There should also be markers for all apiaries in the list. This PR adds that functionality.

Connects #300 

### Demo
screen shot 2018-11-02 at 4 07 46 pm

screen shot 2018-11-02 at 4 07 40 pm

### Notes

Follow up work:
- Map should handle double click zooms vs single clicks to add a marker
- Marker styling should be a circle matching its corresponding card

## Testing Instructions

Click the map and see markers are added. Once a marker is added aka an apiary is added to the state's list of apiaries, the sidebar should change to the apiary card list. They are still dummy cards.

The `main` store also updates! You will need to add `main` to the store whitelist if you want to observe main persisting.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -2,4 +2,4 @@ import { createAction } from 'redux-act';
 
 export const setSort = createAction('Set apiary sort');
 export const setForageRange = createAction('Set forage range');
-export const setApiaryList = createAction('Adds a location to the apiary list');
+export const setApiaryList = createAction('Set the apiary list');
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -2,3 +2,4 @@ import { createAction } from 'redux-act';
 
 export const setSort = createAction('Set apiary sort');
 export const setForageRange = createAction('Set forage range');
+export const addApiary = createAction('Adds a location to the apiary list');
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -2,4 +2,4 @@ import { createAction } from 'redux-act';
 
 export const setSort = createAction('Set apiary sort');
 export const setForageRange = createAction('Set forage range');
-export const addApiary = createAction('Adds a location to the apiary list');
+export const setApiaryList = createAction('Adds a location to the apiary list');
--- a/src/icp/apps/beekeepers/js/src/components/Map.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/Map.jsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { Map as LeafletMap, TileLayer } from 'react-leaflet';
 import { connect } from 'react-redux';
+import { arrayOf, object } from 'prop-types';
 
 import { MAP_CENTER, MAP_ZOOM } from '../constants';
 
-const Map = () => (
-    <div className="map">
-        <LeafletMap
-            center={MAP_CENTER}
-            zoom={MAP_ZOOM}
-        >
-            <TileLayer
-                attribution="&amp;copy <a href=&quot;http://osm.org/copyright&quot;>OpenStreetMap</a> contributors"
-                url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
-            />
-        </LeafletMap>
-    </div>
-);
+class Map extends Component {
+    constructor() {
+        super();
+        this.onClickAddMarker = this.onClickAddMarker.bind(this);
+    }
+
+    onClickAddMarker(event) {
+        window.console.log(event);
+        const { apiaries } = this.props;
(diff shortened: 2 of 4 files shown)


src/icp/apps/beekeepers/js/src/components/Map.jsx
```diff
@@ -1,27 +1,61 @@
...
+        const { apiaries } = this.props;
+        apiaries.push({
+            name: 'Geocoded name',
+            location: event.latlng,
+        });
+        store.dispatch(addApiary(apiaries));
+    }
+
+    render() {
+        return (
```
kellyi: I think you can remove the whole `shouldComponentUpdate` method above and move the marker rendering here, immediately before the `return` statement:

```
render() {
    const { apiaries } = this.props;
    window. console.warn(apiaries); // to check that they change
    const markers = apiaries.map(apiary => (
            <Marker key={`marker-${apiary.name}`} position={[apiary.location.lat, apiary.location.lng]} />
        ));
```

Then in the `return` statement below, replace `{this.markers}` with `{markers}`, which will display the constant just declared.

src/icp/apps/beekeepers/js/src/components/Map.jsx
```diff
@@ -1,27 +1,61 @@
...
+        ));
+    }
+
+    onClickAddMarker(event) {
+        const { apiaries } = this.props;
+        apiaries.push({
+            name: 'Geocoded name',
+            location: event.latlng,
+        });
+        store.dispatch(addApiary(apiaries));
```
kellyi: I'm not sure whether this is affecting it but there are a couple changes to try here:

First, `dispatch` is available as a prop here, so you can just get it from `const { apiaries, dispatch } = this.props;` by virtue of having connected the component.

Second, I think there might be some weirdness going on with the `push` method here -- `push` will mutate the array it's called on which means that it will add the value to the array stored in Redux, then dispatch an action which adds the value again? I think the desired outcome here is to concat a new apiary located at the clicked point to the existing list of apiaries, so either you could:

1. concat the new apiary to the list here, then send the whole list into Redux to concat it to the array
2. send this single apiary to Redux which will do the concat.

Looks like it's been set up to send the whole list in & replace, so I think you can do:

```
        const { apiaries, dispatch } = this.props;
        const newApiary = {
            name: 'Geocoded name',
            location: event.latlng,
        };

        const existingApiariesPlusNewApiary = apiaries.concat(newApiary); // returns a new list
        dispatch(addApiary(existingApiariesPlusNewApiary);
```

rajadain:
Taking a look now.

rajadain:
If this gets in before #317, I can make changes there to make sure this uses the `Apiary` prop type.

fungjj92:
No worries! I think it makes more sense to rebase this PR on #317.

src/icp/apps/beekeepers/js/src/components/Sidebar.jsx
```diff
@@ -12,9 +12,10 @@ const Sidebar = ({ apiaries }) => {
         return <Splash />;
     }
 
-    const apiaryCards = apiaries.map(apiary => (
-        <ApiaryCard key={apiary.marker} apiary={apiary} />
-    ));
+    const apiaryCards = apiaries.map((apiary, idx) => {
+        const key = String.fromCharCode(idx);
+        return <ApiaryCard key={key} apiary={apiary} />;
```
fungjj92: peh, i dont love this but for now to appease React's want of unique keys for change detection. this will need to get replaced with more unique info of the apiary. Since we have a bunch of repeat dummy data we don't have much to work off.

kellyi: Another thing you could do here is to `.join(', ')` on the coordinates list.

fungjj92: that's true, you can't have two points at the same coords

fungjj92: spoke too soon. id need to add a check for that. for now i'll keep this code as this

rajadain:
Taking another look.

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -2,3 +2,4 @@ import { createAction } from 'redux-act';
 
 export const setSort = createAction('Set apiary sort');
 export const setForageRange = createAction('Set forage range');
+export const setApiaryList = createAction('Adds a location to the apiary list');
```
rajadain: Consider renaming the description of this action to "Set the apiary list"

src/icp/apps/beekeepers/js/src/components/Sidebar.jsx
rajadain: I think this solution is good for now. Consider adding a `// TODO` comment to change to something else once the dummy data is removed.

fungjj92:
Thank you for good reviews! I'll make those last suggestions and merge after Jenkins ✅

Issue #316, opened 2018-10-31
Implement Forage Range, Sort selectors


rajadain:
As shown in the mockup:

![image](https://user-images.githubusercontent.com/1430060/47813033-47514300-dd20-11e8-90cd-39c404ad4aca.png)

PR #324, opened 2018-11-07
Implement forage and sort dropdown selectors on sidebar

fungjj92:
## Overview

Continuing to build out the frontend, I converted the HTML dropdowns into react component dropdowns. These dropdowns will change the data and data arrangement in the apiary cards, eventually. Both dropdowns kick off their associated action (see gif) but you will only see data update in the sidebar when toggling the forage range dropdown.

Connects #316 

### Demo

dropdownselectsbees

### Notes

- The styling follows the prototype, which means there's a sllight discrepency with the wireframe
-  I changed the eslint config because the eslint rule the form format was complaining about was satisfied, there are several related issues out there see: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/302
- We don't know what the sorts will be, so for now that dropdown is pretty sparse
- I've updated the s3 bucket and file naming from say '3k' --> '3km' 

## Testing Instructions

Pull down the branch and toggle the dropdowns. See the actions in the console and the data change for forage range toggling.


rajadain:
I think the wireframe is more recent, and we should follow its styling. Can you confirm that, @jfrankl?

jfrankl:
@rajadain yes, the wireframes are more up to date than the prototype. If there are any specific questions about how to style a particular component, feel free to reach out.

fungjj92:
screen shot 2018-11-07 at 2 38 16 pm

Updated

fungjj92:
@jfrankl I truly don't know if I did BEM correctly, particularly as it pertains to the controls block.

jfrankl:
> @jfrankl I truly don't know if I did BEM correctly, particularly as it pertains to the controls block.

No problem! I am happy to take a look. This looks really close. One part that is off is that each CSS selector should contain, at maximum, one block, one element, and one modifier. The format is block__element--modifier. It's okay to not have an element or modifier. There should always be a block.

Your code seems to have two elements. The SCSS transpiles to `.sidebar__header__controls`.

If you unnest the `__controls` element, you'd get `.sidebar__controls`, which is valid BEM.

jfrankl:
Here is a quick BEM naming guide: http://getbem.com/naming/

I'm still learning BEM too. I am happy to talk through any unusual cases. The rest of the design team is available as well.

src/icp/apps/beekeepers/.eslintrc
```diff
@@ -15,6 +15,16 @@
         "react/jsx-indent": [2, 4],
         "react/jsx-indent-props": [2, 4],
         "max-statements": [0, 10],
+        "jsx-a11y/label-has-associated-control": [ "error", {
+            "required": {
+              "some": [ "nesting", "id"  ]
```
rajadain: Incorrect indentation: use 4 spaces.

src/icp/apps/beekeepers/.eslintrc
```diff
@@ -15,6 +15,16 @@
...
+            }
+        }],
+        "jsx-a11y/label-has-for": [ "error", {
+            "required": {
+            "some": [ "nesting", "id"  ]
```
rajadain: Missing indentation: use 4 spaces.

src/icp/apps/beekeepers/.eslintrc
rajadain: Also consider adding some detail in the commit message for why these rules are needed.

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
```diff
@@ -79,6 +85,11 @@ const ApiaryCard = ({ apiary }) => {
 
 ApiaryCard.propTypes = {
     apiary: Apiary.isRequired,
+    forageRange: string.isRequired,
 };
 
-export default ApiaryCard;
+function mapStateToProps(state) {
```
rajadain: Did you consider passing in `forageRange` as a prop from `Sidebar`, rather than connecting this to redux? I find that the fewer connected components there are, the easier it is to reason about an app. I'm not against this decision, just want to make sure it was considered.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -1,5 +1,10 @@
-export const FORAGE_RANGE_3K = '3k';
-export const FORAGE_RANGE_5K = '5k';
+export const FORAGE_RANGE_3KM = '3km';
+export const FORAGE_RANGE_5KM = '5km';
+
+export const FORAGE_RANGES = {
+    '3km': 'threeKm',
```
rajadain: Prefer reusing the constants defined above, to ensure the strings stay in sync:

```js
export const FORAGE_RANGES = {
    [FORAGE_RANGE_3KM]: 'threeKm',
    [FORAGE_RANGE_5KM]: 'fiveKm',
}
```

rajadain: I also have this nagging feeling that having these two representations is going to cause confusion down the line. We should think about using only one string constant per forage range if possible ... but that's a general comment and not necessarily pertinent to this PR.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -15,3 +20,5 @@ export const RASTERS = {
 
 export const MAP_CENTER = [37.899, -97.079];
 export const MAP_ZOOM = 10;
+
+export const SORT_OPTIONS = ['Default', 'Alphabetical'];
```
rajadain: If `SORT_CREATED` from above is not a valid sort option, consider deleting it.

Also, from a comment from @mmcfarland 	in Slack:

> I think the mockup says Default for the sort order, but we'll want each data layer as an option (ie, sort by Nesting score, Pesticide Score, etc)

If these sorts cannot be implemented in this card, make a follow up card for that work.

fungjj92: At the time I put up the PR I didn't know the sort types. I can implement the sorts Matt named.

src/icp/apps/beekeepers/js/src/constants.js
fungjj92: Hmm yeah, good feeling to bring up. While it is a valid solution to numbers in keys (`3km: ...`), I don't love the written version "threeKm" because it's awkward to human readers. Calling a variable in a key by enclosing in `[]` otherwise solves the original issue. For the UI, it is rigid requirement to show numerical values (3km, 5km) so let's go with that.
```
{ ...
    scores: {
        [FORAGE_RANGE_3KM]: { ...}
    }
}
```

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
fungjj92: This is interesting.
In the case of apiary cards, which are created and added to the UI conditionally, I am not sure what the component initialization process looks like with regard to getting data from parent or redux, and need to learn more. My intuition is that passing all data needed is cleaner in this case.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -11,5 +12,7 @@ export const INDICATORS = {
     FORAGE_FALL: 'forage_fall',
 };
 
+export const SORT_OPTIONS = [DEFAULT_SORT].concat(Object.values(INDICATORS));
```
rajadain: I wonder if we'll need to add a [polyfill for `Object.values`](https://www.npmjs.com/package/babel-plugin-transform-es2017-object-entries) which is [not supported in IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values#Browser_compatibility).

fungjj92: I tested in IE11 on browserstack and everything works, no console errors. We have @babel/polyfill installed which I think does it. https://babeljs.io/docs/en/babel-polyfill https://github.com/zloirock/core-js#ecmascript-object

fungjj92:
Ready for re-review! 😪

rajadain:
Taking a look now.

fungjj92:
Thank you for the reviews! Merging.

Issue #320, opened 2018-11-05
Map should handle click scenarios withe more finesse


fungjj92:
As of #319, clicking the map adds a marker for an apiary. The map also zooms on double click. However, the two aren't differentiated and zooming this way drops a pin and adds an apiary to the list.

Assuming we want double-click-zoom functionality, we _don't_ want double clicks to drop a pin.

From #341, clicking the geocoder, or any map control for that matter, should not add an apiary marker either.

fungjj92:
Helpful indeed! Thank you.

Issue #301, opened 2018-10-19
Add Geocoder to Map Component


rajadain:
There should be a geocoder on the map as shown in the wireframes:

![image](https://user-images.githubusercontent.com/1430060/47241463-be4e1980-d3b9-11e8-9c6c-c25b207cf7ba.png)

Requires #300

fungjj92:
Requires #300

rajadain:
"Requires" can't be a comment, you have to edit the main description to include it.

PR #341, opened 2018-11-13
Add ESRI geocoder to map

fungjj92:
## Overview

The user should be able to search for a location on the map. This PR adds the ESRI geocoder, which is Civic Apps' standard geocoder choice. ESRI Geocoder will be helpful in the future when we have to reverse geocode apiary points for the default apiary name.

Connects #301 

### Demo

screen shot 2018-11-13 at 4 25 10 pm

### Notes

I was digging kinda deep trying to clean this console issue. It doesn't break any app functionality. It is generated every time a geosearched address is selected and the map pans. I thought I'd get the PR up before I spent too much time spinning my wheel, when the non-breaking console error could be pinched off for the future or maybe someone is familiar with it.

screen shot 2018-11-13 at 4 24 14 pm

Another known issue I'll add to #320 is that clicking the Geocoder sometimes (often) adds a marker underneath. The map clicking needs more finesse.

## Testing Instructions

Use the geocoder! ~ See the map pans to the location.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/Map.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/Map.jsx
@@ -41,6 +41,13 @@ class Map extends Component {
             expanded: true,
             collapseAfterResult: false,
             position: 'topleft',
+            /*
+                The geocoder and leaflet error without `useMapBounds` and
+                `zoomToResult` set to false.
+                See: https://github.com/Esri/esri-leaflet-geocoder/issues/209
+            */
+            useMapBounds: false,
+            zoomToResult: false,
         }).addTo(map);
 
         geocoder.on('results', ({ results }) => {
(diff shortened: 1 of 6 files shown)


fungjj92:
I investaged, and found a fix for that error! A weird one. See the latest commit for context, I also submitted an issue upstream to esri-leaflet-geocoder because it appears an issue with their code

src/icp/apps/beekeepers/js/src/components/Map.jsx
```diff
@@ -17,6 +23,34 @@ class Map extends Component {
...
+                    url: geocoderUrl,
+                    maxResults: 7,
+                    categories: ['Address', 'Postal'],
+                }),
+            ],
+            placeholder: 'Find location',
+            expanded: true,
+            collapseAfterResult: false,
+            position: 'topleft',
+            useMapBounds: false,
```
rajadain: Consider adding a comment here explaining why we had to use these flags.

Issue #304, opened 2018-10-19
Fetch Scores for Apiaries


rajadain:
- [x] When an apiary is added, trigger request for scores from back-end
- [x] When forage range is changed, trigger request for scores from back-end

fungjj92:
Consider setting up this process to help #338 whereby the workflow is:
adding apiary --> save Apiary to database --> fetch data --> return Apiary from DB

PR #344, opened 2018-11-16
Frontend fetches real data

fungjj92:
## Overview

The fetch endpoint for data already existed for a location/forage range/indicators. The front end has enough structure to request real data now instead of using mock dummy data. This PR hooks up the front and back ends officially and sets up the flow for when the database exists, authentication exists, and we save apiaries there.

I sketched this out as my train of thought. I can explain verbally if this doesn't make sense:
img_20181116_105318

Connects #304 

### Demo

back-front-connect

sort

### Notes

- I based this branch off the Geocoder branch accidentally. I'll fix later or hopefully the other PR gets approved and merged first haha

- If you click outside PA, the requests loop infinitely because the front end doesn't know how to handle server errors from the fetch endpoint. I'll make a followup issue.

## Testing Instructions

Click points within PA and try to use the dropdowns. All functionality should be maintained, plus real data now.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -11,21 +11,20 @@ export const completeFetchApiaryScores = createAction('Completed fetching apiary
 
 export function fetchApiaryScores(apiary, forageRange) {
     return (dispatch, getState) => {
-        dispatch(startFetchApiaryScores());
-
         if (!apiary || !forageRange) {
             const err = !apiary
                 ? 'No apiary provided' : 'No forage range provided';
             window.console.warn(err);
-            return dispatch(failFetchApiaryScores(err));
+            return false;
         }
 
         // check if data for that apiary already exists
         // return the scores and complete the action loop if it exists
-        // TODO?: perform this sooner, perhaps here or in the component
         if (apiary.scores[forageRange].data) {
-            return dispatch(completeFetchApiaryScores());
+            return true;
         }
+
+        dispatch(startFetchApiaryScores());
 
         const {
             main: {
@@ -50,7 +49,7 @@ export function fetchApiaryScores(apiary, forageRange) {
                 // For unauthenticated user relying purely on redux/
                 // localStorage, manually update the apiary listing with data
                 const removedApiaryList = apiaries.filter(a => (
-                    a.location !== apiary.location
+                    !a.location.equals(apiary.location)
                 ));
                 const newList = removedApiaryList.concat(updatedApiary);
 
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -1,4 +1,5 @@
 import { createAction } from 'redux-act';
+import update from 'immutability-helper';
 import axios from 'axios';
 import { INDICATORS } from './constants';
 
@@ -39,8 +40,11 @@ export function fetchApiaryScores(apiary, forageRange) {
                 indicators: Object.values(INDICATORS),
             })
             .then(({ data }) => {
-                const updatedApiary = Object.assign({}, apiary);
-                updatedApiary.scores[forageRange] = data;
+                const updatedApiary = update(apiary, {
+                    scores: {
+                        [forageRange]: { $set: data },
+                    },
+                });
 
                 // TODO: Authenticated user should save apiaryWithData to
                 // the database and update the redux store of apiaries
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -48,6 +48,7 @@ export function fetchApiaryScores(apiaryList, forageRange) {
             }
             return apiary;
         });
+        dispatch(setApiaryList(fetchingApiaries));
 
         const locationList = apiaryList.map(apiary => apiary.location);
 
@@ -65,16 +66,14 @@ export function fetchApiaryScores(apiaryList, forageRange) {
                 // For unauthenticated user relying purely on redux/
                 // localStorage, manually update the apiary listings with data
                 const nonFetchingApiaryList = fetchingApiaries.filter(a => !a.fetching);
-                const apiaryListWithData = apiaryList.map((apiary) => {
-                    const latLngName = String(apiary.location.lat.toFixed(10))
-                        + String(apiary.location.lng.toFixed(10));
-                    return update(apiary, {
+                const apiaryListWithData = apiaryList.map((apiary, idx) => (
+                    update(apiary, {
                         scores: {
-                            [forageRange]: { $set: data[latLngName] },
+                            [forageRange]: { $set: data[idx] },
                         },
                         fetching: { $set: false },
-                    });
-                });
+                    })
+                ));
 
                 const newList = nonFetchingApiaryList.concat(apiaryListWithData);
 
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -11,7 +11,6 @@ export const completeFetchApiaryScores = createAction('Completed fetching apiary
 
 export function fetchApiaryScores(apiary, forageRange) {
     return (dispatch, getState) => {
-        // dispatch initial action declaration
         dispatch(startFetchApiaryScores());
 
         if (!apiary || !forageRange) {
@@ -22,8 +21,8 @@ export function fetchApiaryScores(apiary, forageRange) {
         }
 
         // check if data for that apiary already exists
-        // return the scores and complete action if exists
-        // perform this sooner, perhaps here or in the component
+        // return the scores and complete the action loop if it exists
+        // TODO?: perform this sooner, perhaps here or in the component
         if (apiary.scores[forageRange].data) {
             return dispatch(completeFetchApiaryScores());
         }
@@ -47,7 +46,18 @@ export function fetchApiaryScores(apiary, forageRange) {
                             [forageRange]: data,
                         },
                     });
-                const newList = apiaries.concat(apiaryWithData);
+
+                // TODO: Authenticated user should save apiaryWithData to
+                // the database and update the redux store of apiaries
+                // from a GET/list call
+
+                // For unauthenticated user relying purely on redux/
+                // localStorage, manually update the apiary listing with data
+                const removedApiaryList = apiaries.filter(a => (
+                    a.location !== apiary.location
+                ));
+                const newList = removedApiaryList.concat(apiaryWithData);
+
                 dispatch(completeFetchApiaryScores());
                 dispatch(setApiaryList(newList));
             })
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -1,5 +1,60 @@
 import { createAction } from 'redux-act';
+import axios from 'axios';
+import { INDICATORS } from './constants';
 
 export const setSort = createAction('Set apiary sort');
 export const setForageRange = createAction('Set forage range');
 export const setApiaryList = createAction('Set the apiary list');
+export const startFetchApiaryScores = createAction('Start fetching apiary scores');
+export const failFetchApiaryScores = createAction('Failed to fetch apiary scores');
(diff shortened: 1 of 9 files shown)


rajadain:
Not sure why, but the front-end is fetching data twice when adding a single apiary:



image

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -1,5 +1,60 @@
...
+                ? 'No apiary provided' : 'No forage range provided';
+            window.console.warn(err);
+            return dispatch(failFetchApiaryScores(err));
+        }
+
+        // check if data for that apiary already exists
+        // return the scores and complete action if exists
+        // perform this sooner, perhaps here or in the component
+        if (apiary.scores[forageRange].data) {
+            return dispatch(completeFetchApiaryScores());
```
rajadain: Similar to the above, if the apiary data is already available, simply exit the function, rather than dispatching actions.

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -47,7 +46,18 @@ export function fetchApiaryScores(apiary, forageRange) {
...
-                const newList = apiaries.concat(apiaryWithData);
+
+                // TODO: Authenticated user should save apiaryWithData to
+                // the database and update the redux store of apiaries
+                // from a GET/list call
+
+                // For unauthenticated user relying purely on redux/
+                // localStorage, manually update the apiary listing with data
+                const removedApiaryList = apiaries.filter(a => (
+                    a.location !== apiary.location
```
rajadain: Because `location` is an object, this comparison will always be true:

```js
> node
> {lat: 5, lng: 3} !== {lat: 5, lng: 3}
true
```

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
```diff
@@ -35,6 +35,26 @@ const ApiaryCard = ({ apiary, forageRange }) => {
         return '';
     })();
 
+    if (!apiary.scores[forageRange]) {
+        return (
+            <li className="card">
```
rajadain: Why change this entire block when only the contents of `card__bottom` differ based on Apiary score availability?

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
```diff
@@ -35,7 +37,8 @@ const ApiaryCard = ({ apiary, forageRange }) => {
         return '';
     })();
 
-    if (!apiary.scores[forageRange]) {
+    if (!Object.keys(apiary.scores[forageRange]).length) {
+        dispatch(fetchApiaryScores(apiary, forageRange));
```
rajadain: `dispatch` or otherwise changing state in the `render` method (or a pure functional component) should be avoided. Can this be moved elsewhere?

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
```diff
@@ -4,14 +4,17 @@ import { arrayOf, string } from 'prop-types';
 import { Score } from '../propTypes';
 import { toDashedString, toSpacedString } from '../utils';
 
-const ScoresLabel = ({ indicator, scores }) => (
-    <div className={`indicator indicator--${toDashedString(indicator)}`}>
-        <div className="indicator__number">
-            {scores.map(({ data, error }) => (error ? '!' : data)).join('/')}
+const ScoresLabel = ({ indicator, scores }) => {
+    const score = scores[0] ? scores.map(({ data, error }) => (error ? NaN : Math.round(data))).join('/') : '!';
```
rajadain: This nested tertiary expression is hard to read. Consider breaking into separate named expressions to simplify.

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
fungjj92: Hah, good catch

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -1,5 +1,60 @@
...
+export const startFetchApiaryScores = createAction('Start fetching apiary scores');
+export const failFetchApiaryScores = createAction('Failed to fetch apiary scores');
+export const completeFetchApiaryScores = createAction('Completed fetching apiary scores');
+
+export function fetchApiaryScores(apiary, forageRange) {
+    return (dispatch, getState) => {
+        // dispatch initial action declaration
+        dispatch(startFetchApiaryScores());
+
+        if (!apiary || !forageRange) {
```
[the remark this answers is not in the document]
fungjj92: o true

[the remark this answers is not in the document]
kellyi: For what it's worth, we do periodically use this pattern for client-side validation. The idea is that the `fail...` version of the action can send back a standard-shape error message for display in the UI component by putting it into the Redux store -- and then the UI component can be indifferent about the actual cause of the error except for rendering the message.

Similarly, we periodically do use the `dispatch(complete...` syntax when the data is already available somewhere in the client and doesn't require an API request to get it. In those cases, we often pluck the data object out of whatever container in order to send it through the `completeFetch....` machinery so that it gets to the proper place in Redux -- which also allows the UI component to be relatively indifferent to how the data's been gotten.

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -1,5 +1,60 @@
...
+        } = getState();
+
+        return axios
+            .post('/beekeepers/fetch/', {
+                location: apiary.location,
+                forage_range: forageRange,
+                indicators: Object.values(INDICATORS),
+            })
+            .then(({ data }) => {
+                const apiaryWithData = Object.assign(apiary,
```
kellyi: Just a heads up that `Object.assign` will mutate its first argument, which looks like it's coming from the Redux store. To avoid that you can use an empty object as the first arg -- `Object.assign({}, apiary, ....` -- or use `immutability-helper` like:

```
update(apiary, {
   scores: {
       [forageRange]: { $set data },
   },
}),

File: src/icp/apps/beekeepers/js/src/actions.js (hunk 6 above)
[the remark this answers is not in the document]
Comment (fungjj92) [reply src/icp/apps/beekeepers/js/src/actions.js]: Neglected to mention I lifted this fetch pattern from American Water. I liked the organized logging when I read the AmWater code but also see that it is a little misleading for debugging that an action can have more than 1 meaning. Like, failure meaning failure to provide params to `fetchApiaryScores` or failure from the backend endpoint to query s3.

File: src/icp/apps/beekeepers/js/src/actions.js (hunk 7 above)
Comment (fungjj92) [reply src/icp/apps/beekeepers/js/src/actions.js]: Oh right on. I hadn't thought of using `immutability-helper` outside of the reducers. I like the readability of that code.

**Comment by fungjj92** [comment]
This is ready for review :) CI is failing, I'll investigate, but I don't think that will interfere with review~

**Comment by rajadain** [comment]
Taking another look now!

File: src/icp/apps/beekeepers/js/src/actions.js [hunk 8]
```diff
@@ -1,5 +1,98 @@
...
+        } = getState();
+
+        // finally, declare starting to fetch scores
+        dispatch(startFetchApiaryScores());
+
+        // mark apiaries from input list as fetching data
+        const fetchingApiaries = apiaries.map((apiary) => {
+            if (apiaryList.find(a => Object.is(a.location, apiary.location))) {
+                return update(apiary, {
+                    fetching: { $set: true },
```
rajadain: By updating the global state in the action and not the reducer, we are breaking the redux pattern here. Consider setting `fetching` to `true` in `reducers.js` as a response to `startFetchApiaryScores`.

This method should assemble data (such as find out which apiaries need to be fetched, and what scores have been returned), then dispatch actions with that data as parameters, and the reducer should update the redux state. If the redux state is mutated outside the reducer cycle, that makes the app much harder to reason about, because there would be multiple places where the state is touched, and we won't see them in the redux log in the console. Does that make sense?

fungjj92:
Good observation, I should have made a note about using `lat` and `lon` as an identifying key. Since the fetch data endpoint accepts a list of locations now, I needed a reliable way to key locations to their apiary. There's no unique information to go off of to identify an `Apiary` at the moment besides latLon. When we assign unique markers, names, or IDs/PKs, we will have a better naming system to work off.

rajadain:
One thing that could be used to identify the locations is to send back the responses in an array in the same order as the input array. So the input array index becomes the identifier.

src/icp/apps/beekeepers/js/src/actions.js
fungjj92: What I realized and learned from an earlier comment by @kellyi is that the `update` method from `immutability-helper` has nothing to do with updating the state but rather is a helper that handles deep copying and reassignment of an object, thus preventing mutation. I'd only seen `update` used in reducers and had conflated state updating with `update`. Does this address the confusion?

What I did forget to do here, though, is dispatching the new apiary list after updating the `fetching` properties to `true`. See below:

screen shot 2018-11-26 at 1 51 50 pm

kellyi: Appears that the Redux state currently stores the apiaries in a list, and each list items is responsible for handling its own fetching, data, and error states. I think that makes sense to me, but/and it also adds a minor bit of complexity since the app now has to find the data fetching object by array index rather than by key.

Seems like, generally, the `start`, `complete`, and `fail` fetch things here now each need to send back a full, albeit transformed copy of the apiary under consideration back to the reducer. The reducer, in turn, needs to use something like `findIndex` to find the index of the specific apiary, then do something like

```
const indexOfApiaryToUpdate = apiaries.findIndex(.......whatever function => updatedApiary);
...
apiaries: { $set: [
   ....state.apiaries.slice(0, indexOfApiaryToUpdate),
   updatedApiary,
   ...state.apiaries.slice(indexOfApiaryToUpdate + 1),
], },
```

If it's helpful: in place of the inline reducer functions we usually use with redux-act, you can also write more complicated functions that have a signature of `(state, payload) => state`, which will just work as long as they return a new state. So you could do something like

```
[startFetchApiaryData]: handleStartFetchApiaryData,
```

&

```
const handleStartFetchApiaryData = (state, payload) => {
   ....array splicing, immutability updates, etc ...
}
```

fungjj92: Perhaps the comment's location is confusing me. Can you refer to a particular line? The code updates the state with a reducer, i.e. calling `dispatch` on `setApiaryList`. No where else in this function is the state being updated outside a reducer as far as I understand?

rajadain: Ah! You're right! I was making the same mis-assumption that `update` === state change. I see that this is only being used for assigning values to `fetchingApiaries` here. Sorry about that, please disregard this thread.

fungjj92: Didn't see Kelly comment until I after I posted mine. Reading.

kellyi: > Didn't see Kelly comment until I after I posted mine. Reading.

FWIW, the way it's done here does seem to avoid mutating the Redux store -- it's already creating the new list. So I think the main concern is whether it feels more complicated to do that list manipulation here or in whatever reducers.js file.

Either way is probably fine -- just seems like a question of where you'd prefer the complexity to live.

rajadain: I would generally prefer complexity to live in `actions.js`, as done here, rather than `reducers.js`.

fungjj92: @kellyi  Ah, interesting re: reducers accepting functions. I suppose I could have expected that. I think, as you point out, the complexity would be about the same but just in a different location. A reducer to update individual apiaries in the list would be noisier, so I think I will leave this code as is.

Yay this resulted in a very educational thread. Resolving.

fungjj92:
I was nervous about relying on array ordering alone, hence trying to come up with a unique key system, but reliable ordering is part of arrays' whole purpose. Something to be leaned on. As suggested, I converted to using arrays throughout the data fetch and update apiaries process.

fungjj92:
Thanks for the rigorous reviews !

Issue #334, opened 2018-11-13
Make Django Model for Survey


rajadain:
Fields:

- `monthyear`: string
- `created_at`: datetime
- `numcolonies`: int
- `apiary_id`

Constraints:

- An Apiary can have zero or more number of Surveys.
- A Survey belongs to exactly one Apiary.
- The combination of `apiary_id` and `monthyear` should be unique.
- A Survey can have at most 3 Subsurveys

PR #347, opened 2018-11-27
Add Beekeepers Survey Model

rajadain:
## Overview

Adds model for Survey.

Connects #334 

### Notes

Includes an additional `surveytype` field not specified in the original card, since now we have to support more than one kind of survey: a long one and a short one, as discussed in [this document](https://docs.google.com/document/d/1WBEbPIAoex_kRQydom1420s_gBrEAk9Sh2sByYFHFjE/edit), the final details are still being worked out.

## Testing Instructions

* Check out this branch and run `manage.sh migrate`
* Ensure the table is created appropriately

### DIFF
--- a/src/icp/apps/beekeepers/models.py
+++ b/src/icp/apps/beekeepers/models.py
@@ -47,3 +47,42 @@ class Apiary(models.Model):
 
     def __unicode__(self):
         return unicode('{}:{}'.format(self.user.username, self.name))
+
+
+class Survey(models.Model):
+    """Survey of questions for an Apiary for a given month"""
+    APRIL = 'APRIL'
+    NOVEMBER = 'NOVEMBER'
+    MONTHLY = 'MONTHLY'
+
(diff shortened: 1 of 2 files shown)


src/icp/apps/beekeepers/models.py
```diff
@@ -47,3 +47,37 @@ class Apiary(models.Model):
 
     def __unicode__(self):
         return unicode('{}:{}'.format(self.user.username, self.name))
+
+
+class Survey(models.Model):
+    """Survey of questions for an Apiary for a given month"""
+    SURVEY_TYPES = (
+        ('LONG', 'Long'),
```
fungjj92: Will we ever need to refer to these options individually ? If so, consider another format for them (Enum, variables, etc. I tried out different formats for #350 though I'm not 100% on my choice there)

rajadain:
Good suggestions! Updated.

Issue #346, opened 2018-11-26
Investigate Build RAM Issues


rajadain:
Often times, running `beekeepers build` after starting the VM results in out of memory errors. This is observed both on local developer machines and on CI for PR builds. The disparity between webpack's `build` and `dev-server` behavior has been talked about elsewhere, e.g. https://github.com/webpack/webpack/issues/6389 and https://blog.box.com/blog/how-we-improved-webpack-build-performance-95.

Figure out what is causing the excessive RAM usage, and address the issue.

Another fix may be to increase the RAM for the `app` VM.
Another fix may be to add a swapfile with https://github.com/azavea/ansible-swapfile.

PR #349, opened 2018-11-28
Add swapfile to application server

hectcastro:
## Overview

During Ansible provisioning runs, the `Install NPM Dependencies` Beekeepers task occasionally causes the Node.js process to run out of memory. In an effort to allow this one-off process to have more memory to operate, a swapfile was added to boost virtual memory.

Fixes https://github.com/project-icp/bee-pollinator-app/issues/346

## Testing Instructions

Review the contents of the following PR build logs:

- http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-pull-requests/205/
- http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-pull-requests/206/
- http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-pull-requests/208/
- http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-pull-requests/209/
- http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-pull-requests/210/

Also, you can apply the changes in this branch to your `app` virtual machine and determine if swap is enabled:

```bash
$ vagrant up --provision app
$ vagrant ssh app
vagrant@app:~$ sudo swapon -s
Filename                                Type            Size    Used    Priority
/swapfile                               file            976556  0       -1
```

### DIFF
--- a/deployment/ansible/roles.yml
+++ b/deployment/ansible/roles.yml
@@ -81,3 +81,6 @@
 
 - src: azavea.memcached
   version: 0.1.0
+
+- src: azavea.swapfile
+  version: 0.1.0
--- a/deployment/ansible/roles/bee-pollinator.app/meta/main.yml
+++ b/deployment/ansible/roles/bee-pollinator.app/meta/main.yml
@@ -1,6 +1,7 @@
 ---
 dependencies:
   - { role: "bee-pollinator.base" }
+  - { role: "azavea.swapfile" }
   - { role: "azavea.python", python_development: True }
   - { role: "azavea.pip" }
   - { role: "azavea.nodejs" }


hectcastro:
@azavea-bot rebuild

hectcastro:
@azavea-bot rebuild

hectcastro:
I don't have a good answer for that question. My suspicion would be similar: whatever algorithm governs virtual memory for the OS within the VM is now incorporating the swap into its decision-making process, which is occasionally causing it to get allocated to other tasks too (instead of more aggressive deallocation?).

cc: @jamesmcclain

jamesmcclain:
I don't have a good answer for that question. My suspicion would be similar: whatever algorithm governs virtual memory for the OS within the VM is now incorporating the swap into its decision-making process, which is occasionally causing it to get allocated to other tasks too (instead of more aggressive deallocation?).

Yes, I believe that is so.  All things being equal, any normal process is a likely to use swap as any other.  It seems to be possible to [configure global swapiness](https://askubuntu.com/questions/103915/how-do-i-configure-swappiness) and [per-cgroup swapiness](https://unix.stackexchange.com/questions/10214/how-to-set-per-process-swapiness-for-linux) if the default behavior is not what is desired.

Issue #290, opened 2018-10-09
Update User Model to Differentiate Beekeepers Users from Pollination Mapper Users


rajadain:
Update the User Model to add a field which says if the user was created from the Beekeepers app or the Pollination Mapper app.

PR #350, opened 2018-11-28
Record through which app User was created

fungjj92:
## Overview

The project already has Users for the Pollination Mapper app. Users from PM and Beekeepers will use the same User table, but we want to record from which app the User signed up. This PR adds a UserProfile attached to the User that can be extended to store other info (such as the future Beekeeper account profile).

This was done in place of moving to a custom `User` model. Django is protective of its User class and adding a custom User model would require copying all data to the new table (can't just add a new column to User). Since Bees is already in production, we want to preserve data / touch it as little as possible. We also gain flexibility to add user info later.

Connects #290

### Demo

After creating a user via the Pollination Mapper sign up modal/workflow:

screen shot 2018-11-28 at 3 52 43 pm
screen shot 2018-11-28 at 3 52 56 pm

## Testing Instructions

Apply migrations `./scripts/manage.py migrate`
Check that your existing users were given userprofiles marked with Pollination as their origin app (all users to this point should originate from there) http://localhost:8000/admin
Create a user via the sign up work flow on http://localhost:8000/ and see that it too assigned a user profile either checking via shell or the admin

### DIFF
--- a/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
+++ b/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
@@ -16,7 +16,7 @@ def create_user_profiles_for_existing_users(apps, schema_editor):
 class Migration(migrations.Migration):
     
     dependencies = [
-        ('user', '0001_initial')
+        ('user', '0001_create_user_profile_model')
     ]
 
     operations = [
--- a/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
+++ b/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
@@ -4,13 +4,13 @@ from __future__ import unicode_literals
 from django.db import migrations, models
 import django.contrib.postgres.fields
 from django.conf import settings
+from django.contrib.auth.models import User
+
+from apps.user.models import UserProfile
 
 def create_user_profiles_for_existing_users(apps, schema_editor):
-    User = apps.get_model('auth', 'User')
-    UserProfile = apps.get_model('user', 'UserProfile')
-
     for user in User.objects.all():
-        UserProfile.objects.create(user=user)
+        UserProfile.objects.get_or_create(user=user, origin_app=UserProfile.POLLINATION)
  
 
 class Migration(migrations.Migration):
--- a/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
+++ b/src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.contrib.postgres.fields
+from django.conf import settings
+
+def create_user_profiles_for_existing_users(apps, schema_editor):
+    User = apps.get_model('auth', 'User')
+    UserProfile = apps.get_model('user', 'UserProfile')
+
+    for user in User.objects.all():
+        UserProfile.objects.create(user=user)
+ 
+
+class Migration(migrations.Migration):
+    
+    dependencies = [
+        ('user', '0001_initial')
+    ]
+
+    operations = [
+        migrations.RunPython(create_user_profiles_for_existing_users)
+    ]
--- a/src/icp/apps/user/models.py
+++ b/src/icp/apps/user/models.py
@@ -1,1 +1,25 @@
 # -*- coding: utf-8 -*-
+from django.db import models
+from django.contrib.postgres.fields import ArrayField
+from django.conf import settings
+
+AUTH_USER_MODEL = getattr(settings, 'AUTH_USER_MODEL', 'auth.User')
+
+
+class UserProfile(models.Model):
+    """For additional user properties."""
+
+    # App options
+    POLL = ('POLL', 'Pollinator')
+    BEEK = ('BEEK', 'Beekeepers')
+    APP_CHOICES = (POLL, BEEK)
+
+    user = models.OneToOneField(AUTH_USER_MODEL)
+    app = ArrayField(
+        models.CharField(
+            max_length=255,
+            choices=APP_CHOICES,
+        ),
+        default=[POLL],
+        blank=True,
+    )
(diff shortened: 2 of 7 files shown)


rajadain:
Taking a look now!

src/icp/apps/user/views.py
```diff
@@ -117,6 +118,14 @@ def sign_up(request):
 
     if form.is_valid():
         user = view.register(request, form)
+
+        # Create a user profile with correct origin app
+        # TODO: Correct the check for the beekeepers app
+        if request.META['HTTP_REFERER'] is 'http://localhost:8000/?beekeepers':
+            UserProfile.objects.create(user=user, app=[UserProfile.BEEKEEPERS])
+        else:
+            UserProfile.objects.create(user=user)
```
rajadain: Rather than relying on the default, prefer specifying the app explicitly here, and in the migration.

src/icp/apps/user/models.py
```diff
@@ -1 +1,25 @@
...
+
+    POLLINATORS = ('POLLINATORS', 'Pollinators')
+    BEEKEEPERS = ('BEEKEEPERS', 'Beekeepers')
+    APP_CHOICES = (POLLINATORS, BEEKEEPERS,)
+
+    user = models.OneToOneField(AUTH_USER_MODEL)
+    origin_app = models.CharField(
+        max_length=255,
+        choices=APP_CHOICES,
+        default=POLLINATORS,
```
rajadain: Because the default value is a tuple, not a string, it leads to database entries like this:

```sql
SELECT * FROM user_userprofile;

 id |           origin_app           | user_id
----+--------------------------------+---------
  1 | ('POLLINATORS', 'Pollinators') |       1
(1 row)
```

Whereas we probably want this:

```
 id | origin_app  | user_id
----+-------------+---------
  1 | POLLINATORS |       1
```

Take a look at [how the defaults are specified here](https://github.com/WikiWatershed/model-my-watershed/blob/caf29ee9fb92ea4d9e5ff9eb8be5cd2f77a7df09/src/mmw/apps/user/models.py#L44-L87).

fungjj92: Ah, that's what I was *trying* to go for, but `CHOICES` required tuples, which tripped me up.

src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations, models
+import django.contrib.postgres.fields
+from django.conf import settings
+
+def create_user_profiles_for_existing_users(apps, schema_editor):
+    User = apps.get_model('auth', 'User')
+    UserProfile = apps.get_model('user', 'UserProfile')
+
+    for user in User.objects.all():
+        UserProfile.objects.create(user=user, origin_app=UserProfile.POLLINATION)
```
rajadain: Hmm, when trying to run the migration I get this:

```
Applying user.0002_add_userprofiles_to_existing_users...Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/migration.py", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/special.py", line 183, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/opt/app/apps/user/migrations/0002_add_userprofiles_to_existing_users.py", line 13, in create_user_profiles_for_existing_users
    UserProfile.objects.create(user=user, origin_app=UserProfile.POLLINATION)
AttributeError: type object 'UserProfile' has no attribute 'POLLINATION'
```

Unsure why it doesn't work here.

kellyi: Not sure if this is relevant but I remember encountering an issue in GARP that some kinds of changes to the User model had to occur in migration 1 or 0.

fungjj92: Bringing in `UserProfile` via import statement `from apps.user.models import UserProfile` gives access to the model's variable data whereas the `apps.get_model()` call does not.

src/icp/apps/user/migrations/0002_add_userprofiles_to_existing_users.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.contrib.postgres.fields
+from django.conf import settings
+from django.contrib.auth.models import User
+
+from apps.user.models import UserProfile
```
fungjj92: Alright, to make this work (access variables on the UserProfile model as well as create new UserProfiles with Users) I had to directly import the models.

fungjj92:
Not sure if my last comment was seen, but this should be ready for review.

rajadain:
I'll take another look shortly

fungjj92:
Thanks for reviews!!

Issue #325, opened 2018-11-13
Make the "Participate in study" button in the top-right open a Login / Sign Up modal


fungjj92:
Currently in the app you will see a functionless "Account" button in the top right of the header. First, update that to the purpleish "Participate in study" button. Upon clicking, a modal like that seen below should pop up that can toggle between two forms -- sign up or login.

![image](https://user-images.githubusercontent.com/10568752/48427304-74e1b780-e736-11e8-93eb-a46f68d91d2d.png)

<img width="408" alt="screen shot 2018-12-03 at 11 14 11 am" src="https://user-images.githubusercontent.com/10568752/49386006-a2de6a00-f6ec-11e8-8e6b-8da8bf60dd9a.png">

<img width="212" alt="screen shot 2018-12-03 at 11 14 26 am" src="https://user-images.githubusercontent.com/10568752/49386007-a2de6a00-f6ec-11e8-8abf-198d5d5ad459.png">

kellyi:
In case it helps this is a pretty nice library we've used for modals: https://www.npmjs.com/package/reactjs-popup

There's also react-modal, but I think reactjs-popup has a slightly nicer API.

PR #354, opened 2018-12-03
Implement modals for auth in the UI

fungjj92:
## Overview

We've started working on the user management feature for Beekeepers as of #290 and this PR is the first of many for the UI. Set up the 3 types of auth modals and moving between them as per the wireframes, see issue for screen caps.

Note and pardon the `=` in the branch name, sloppy typing 😬 

Connects #325

### Demo
screen shot 2018-12-03 at 4 07 00 pm

screen shot 2018-12-03 at 4 08 04 pm

screen shot 2018-12-03 at 4 08 13 pm

### Notes

The dispatched action from the modal closing is called twice, which is non-breaking and seems out of my control using the lib's close methods 🤔 -- reactjs-popup appears to be doing that. 

Unsure about my BEM implementation 😬 

## Testing Instructions

Install the new npm lib with `./scripts/beekeepers.sh install`
Click "participate in study" in the top right and cycle through making sure things are as you expect. Some buttons that will eventually have functionality are set up to close their modal at the moment.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/App.jsx
+++ b/src/icp/apps/beekeepers/js/src/App.jsx
@@ -4,11 +4,15 @@ import { hot } from 'react-hot-loader';
 import Header from './components/Header';
 import Map from './components/Map';
 import Sidebar from './components/Sidebar';
+import SignUpModal from './components/SignUpModal';
+import LoginModal from './components/LoginModal';
 
 const App = () => (
     <>
         <Header />
         <main>
+            <SignUpModal />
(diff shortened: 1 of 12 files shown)


rajadain:
whoops sorry, accidentally merged develop onto this branch and pushed it up. Going to undo that.

fungjj92:
Thanks for looking! Made changes as suggested. Further changes can be captured in a styling pass.

screen shot 2018-12-10 at 11 02 25 am

Issue #351, opened 2018-11-29
Add Spinners to Apiary Cards


rajadain:
As Apiary Cards wait for scores, show a spinner instead of the current placeholder message.

Issue #327, opened 2018-11-13
Implement Login workflow


fungjj92:
Allow users to log in with their account credentials via the modal implemented in #325 and update the client state.

Note, there is also a new "Log in" button in the header that should open the login modal.

![image](https://user-images.githubusercontent.com/10568752/48427331-83c86a00-e736-11e8-814a-4b3f0ad1730e.png)

Requires #290, #325

Issue #356, opened 2018-12-04
Handle server error on fetch request 


fungjj92:
The fetch request doesn't gracefully handle a server failure. Rather than send back a list of Apiaries with any error messages handled appropriately in scores, a failure bubbles up to fail the entire request. The UI shows an infinite spinner:

<img width="893" alt="screen shot 2018-12-04 at 12 54 00 pm" src="https://user-images.githubusercontent.com/10568752/49462510-36369e80-f7c4-11e8-999f-9b6d63343583.png">

PR #357, opened 2018-12-04
Add spinner to apiary cards

fungjj92:
## Overview

Replace the TODO text in the apiary card with a spinner. The user sees the spinner when the data is fetching. I lifted the spinner CSS from MMW.

Connects #351 

### Demo

spinner

## Notes
FYI if you’re logged into the pollination front end, beekeepers requests to the backend will fail with a 403 due to CSRF protection. The underlying error: `CSRF Failed: CSRF token missing or incorrect.` It's ok for now, but we will want to fix that after login is implemented #327 . I've created #358

## Testing Instructions
Click on the map, and see the spinner go on the apiary card ~


fungjj92:
Thx for review! I did make a card for that behavior #356 actually

Issue #358, opened 2018-12-04
Attach User's CSRF token to non-GET requests to Django


fungjj92:
If you’re logged into the pollination front end, beekeepers requests to the backend will fail with a `403`. The underlying error: `CSRF Failed: CSRF token missing or incorrect.` To fix, user's non-GET requests to the backend should send their CSRF token too. [Revelant Stackoverflow explanation](https://stackoverflow.com/questions/32653518/django-rest-framework-returning-403-response-on-post-put-delete-despite-allowa).

Discovered locally and on staging.

Requires #327

rajadain:
Shares points with #327

PR #360, opened 2018-12-07
Implement logging in in Beekeepers UI

fungjj92:
## Overview

This PR officially hooks up the beekeepers UI and Django user management. Users registered to either app will be able to sign in.

I added some basic error handling to show the user the error message from the server.

I had to loop in #358 to make a the POST request to `/user/login`. I spent a solid chunk of time solving it then realized GARP had to deal with the same issue, so I lifted that project's solution which is just `axios` set up to recognize available CSRF info.

Connects #358, #327

### Demo

screen shot 2018-12-07 at 12 09 32 pm

sucessfully logged in -->
screen shot 2018-12-07 at 12 09 13 pm

### Notes

As with the other actions fired by the modal's `close` method, clearing the auth error fires twice in most cases too.

I added a reducer set for storing user information. The fields we save are flexible. `username`, `userId`, and `authError` seemed like a good starting set.

My gut tells me using an observable/ listening to the dispatched `login` action is how I really do want to receive any login error but I thought this solution was simple and self-contained enough to prefer the time savings over learning how to do the former. 

Again, BEM... I'm learning 🤷‍♀️ 

## Testing Instructions

You must have a valid user (either thru the django shell or through the pollinators app).
Click "login" and submit your information *incorrectly*. See the error message shows, and doesn't reappear if you exit/reopen the login modal.
Click "login" and submit your info *correctly* and see that login works

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
@@ -4,19 +4,13 @@ import Popup from 'reactjs-popup';
 import { connect } from 'react-redux';
 
 import { closeLoginModal, openSignUpModal, login } from '../actions';
+import { parseFormToObject } from '../utils';
 
 
 const LoginModal = ({ isLoginModalOpen, dispatch }) => {
     const submitForm = (event) => {
-        let form = {};
-        let i;
-        for (i = 0; i < event.currentTarget.elements.length; i += 1) {
-            const inputValue = event.currentTarget.elements[i].value;
-            const inputName = event.currentTarget.elements[i].name;
-            const newElement = { [inputName]: inputValue };
-            form = Object.assign(newElement, form);
-        }
-        dispatch(login(form));
+        // Send form data to backend for validation and prevent modal from closing
+        dispatch(login(parseFormToObject(event)));
         event.preventDefault();
     };
 
--- a/src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { bool, func, string } from 'prop-types';
 import Popup from 'reactjs-popup';
 import { connect } from 'react-redux';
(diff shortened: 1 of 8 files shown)


rajadain:
Meta comment: you have to use one "Connects" per line for Waffle to pick it up, e.g.

Connects #327
Connects #358

src/icp/apps/beekeepers/js/src/utils.js
```diff
@@ -12,3 +12,15 @@ export function toTitleCase(value) {
         .map(word => word.charAt(0).toUpperCase() + word.slice(1))
         .join('');
 }
+
+export function parseFormToObject(submitEvent) {
+    let form = {};
+    let i;
+    for (i = 0; i < submitEvent.currentTarget.elements.length; i += 1) {
```
kellyi: Consider storing these values either in component state or Redux state instead of reading them from the DOM.

src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
```diff
@@ -1,65 +1,116 @@
...
+    return (
+        <Popup
+            open={isLoginModalOpen}
+            onClose={() => {
+                dispatch(clearAuthError());
+                dispatch(closeLoginModal());
+            }}
+            modal
+        >
+            {close => (
```
kellyi: If you drop `close =>` here, pass the inset stuff as a node rather than a function, and then have the

```
() => {
   dispatch(clearAuthError());
   ...
```

...function get called in the button's `onClick` method, does that get rid of the double dispatch error? I think this component's may currently be having two close events: one from it's built-in state, another from the Redux state being updated to close.

In other words, the functions in `onClose` are getting fired both as a response to the component's close method and the `dispatch(closeLoginModal())` call.

rajadain:
This can target `develop` now. Also heads up that there are merge conflicts.

src/icp/apps/beekeepers/js/src/utils.js
fungjj92: A ha! Thanks for this direction. Makes a lot of sense, and feels cleaner. I implemented and it works well with the reactjs-popup library as used. Will push up changes soon.

src/icp/apps/beekeepers/js/src/components/LoginModal.jsx
fungjj92: Nice try! We also need to fire the clean up action(s) in the case where the user quits the modal by clicking the overlay (esc button, or click) which is why I went with the current implementation.

fungjj92:
Merging now.

Issue #329, opened 2018-11-13
Add tabs to the header


fungjj92:
In newest wireframes, the survey and map functionality are split into two different views on the app, accessed via tabs in the header. Add the "Location Finder" and "Survey" tabs as well as the functionality to tab between them. This design should work for the About and Methodology tabs too (when those get pages). Include the empty state for the Survey tab. The default tab is Location Finder.

![image](https://user-images.githubusercontent.com/10568752/48421796-32ff4400-e72b-11e8-8645-3e12de4ed4cc.png)

PR #363, opened 2018-12-11
Add survey route and navigation tabs

fungjj92:
## Overview

The survey view will live at its own route / page and this PR sets up the app to accomodate that future work.

Connects #329

### Demo

screen shot 2018-12-11 at 12 10 47 pm
screen shot 2018-12-11 at 12 10 54 pm

### Notes

The routes don't include the `?beekeepers` query param, which navigates correctly and eventually for staging/production when rather than check for that flag django routes to Beekeepers by the host path (whatever becomes the beekeepers domains). However, for development it is maybe a little annoying because if one refreshes the page they get rerouted to Pollination Mapper.

## Testing Instructions

`./scripts/beekeepers.sh start` and navigate the tabs in the header
Routes at http://localhost:8000/survey/?beekeepers and http://localhost:8000/?beekeepers

### DIFF
--- a/src/icp/apps/beekeepers/js/src/main.jsx
+++ b/src/icp/apps/beekeepers/js/src/main.jsx
@@ -1,12 +1,13 @@
 import React from 'react';
 import { render } from 'react-dom';
 
-import { BrowserRouter, Route } from 'react-router-dom';
+import { BrowserRouter, Route, Switch } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
 
 import '../../sass/main.scss';
 import App from './App';
+import Surveys from './Surveys';
 
 import { store, persistor } from './store';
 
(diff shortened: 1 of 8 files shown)


src/icp/apps/beekeepers/js/src/main.jsx
```diff
@@ -16,7 +17,10 @@ render(
             https://github.com/project-icp/bee-pollinator-app/issues/314 */}
         <PersistGate loading={null} persistor={persistor}>
             <BrowserRouter>
-                <Route exact path="/" component={App} />
+                <Switch>
+                    <Route exact path="/" component={App} />
+                    <Route path="/survey" component={Survey} />
```
rajadain: Given that the only real difference between `App` and `Survey` is the `Map` and `Sidebar` components (and whatever comes in the future within `Survey`, perhaps we should have a shared parent component that shows either the Map + Sidebar / Survey based on the route. That will allow us to do app-level things, like logging in, as done in #362, in one place.

fungjj92: Good input! Fixed up in d2247b1. Works great.

src/icp/apps/beekeepers/js/src/main.jsx
```diff
@@ -16,7 +21,18 @@ render(
             https://github.com/project-icp/bee-pollinator-app/issues/314 */}
         <PersistGate loading={null} persistor={persistor}>
             <BrowserRouter>
-                <Route exact path="/" component={App} />
+                <>
```
fungjj92: Router only wants 1 child, hence nesting the components in a React Fragment

rajadain:
Let's update this to move the `Switch` into `App` as discussed in person this afternoon. You may also consider rebasing this on `develop`.

fungjj92:
👏 team programming 👏 
~ merging ~

Issue #336, opened 2018-11-13
Make Django Model for Subsurvey


rajadain:
Fields:

There should be a column per answer for each question in the [Pro Monthly Survey](https://docs.google.com/document/d/1WBEbPIAoex_kRQydom1420s_gBrEAk9Sh2sByYFHFjE/).

Constraints:

- A Survey must have at least one Subsurvey.
- A Subsurvey will belong to exactly one Survey.

rajadain:
Blocked until the survey fields have been finalized https://docs.google.com/document/d/1WBEbPIAoex_kRQydom1420s_gBrEAk9Sh2sByYFHFjE/edit

rajadain:
Here's the gist:

* There will be two kinds of users: tentatively called Light and Pro
* There will be four kinds of surveys:
  - User Profile survey: will be filled out by all users after they sign up
  - April Survey: a short survey with only a few questions to be filled out in April by all users
  - November Survey: a long survey with a number of questions to be filled out in November by all users
  - Monthly Survey: a long survey with a number of questions to be filled out every month by Pro users
* The questions can be multiple choice, numeric, or text answers
* The final listing can be found here: https://docs.google.com/document/d/1WBEbPIAoex_kRQydom1420s_gBrEAk9Sh2sByYFHFjE/edit

rajadain:
I've made a schema spreadsheet for the four surveys: https://docs.google.com/spreadsheets/d/1lkXvBpeyg-lhzH48CdBe_C-HuepJBaXUM2904umd7PI. There are some questions noted in the "Notes" column that need to be answered before we can proceed, so marking this card as "blocked" again.

rajadain:
A number of the questions in the spreadsheet above have been answered, unblocking this card.

rajadain:
Also add the User Survey model, as a new model so as to not burden Pollination Mapper users.

PR #364, opened 2018-12-11
Add Subsurvey Models

rajadain:
## Overview

Converts the [schema of subsurveys](https://docs.google.com/spreadsheets/d/1lkXvBpeyg-lhzH48CdBe_C-HuepJBaXUM2904umd7PI/) into Django models.

Connects #336 

## Testing Instructions

* Run migrations
* Inspect the generated tables

### DIFF
--- a/src/icp/apps/beekeepers/admin.py
+++ b/src/icp/apps/beekeepers/admin.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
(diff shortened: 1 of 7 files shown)


rajadain:
Added to Django Admin:

image

For **3** above, the answer is that it allows a set of predefined inputs, but also free text input. The way we will handle this is by using the keys listed in the comments for the predefined inputs, and answers prefixed with `OTHER-` for the free text inputs. Since many of these fields support multiple choice answers, all the keys will be joined with `;`.

Issue #342, opened 2018-11-16
Create VRT for multi-state raster sets


mmcfarland:
We'll be supporting PA and IL datasets at launch.  Future versions of the app may add additional state datasets.  To support this, create a VRT for each dataset which includes both state versions.

For example, `pa_nesting_3k.tif` + `il_nesting_3k.tif` = `nesting_3k.tif`

Checklist:
* [x] Convert all 3k datasets to VRT
* [x] Convert all 5k datasets to VRT
* [x] Update apiary score endpoint to point to new file extension

PR #365, opened 2018-12-11
Update rasters for multi-state support

mmcfarland:
## Overview

Rasters are now provided as discrete files representing single states.
In order to mosaic them on the fly and read from the correct file, VRTs
have been created which reference adjacent raw geotiffs. Since rasterio
delegates vsis3 calls directly to gdal for the path, additional work
must be done to provide AWS credentials via environment variables as
rasterio will only continue to read the initial VRT read with it's own
configured credentials.

Connects #342 

### Demo

VRTs and tifs together in a temporary s3 bucket:

screenshot from 2018-12-11 17 29 16

screenshot from 2018-12-11 17 35 12
screenshot from 2018-12-11 17 34 45
screenshot from 2018-12-11 17 33 50

### Notes

Since other devs are currently using the tifs in the real staging bucket, I've created a temp bucket to hold these values until it's ready to be merged.  I'll then copy between buckets.  The bit of code around GDAL credentials was learned the hard way around previous work on USACE doing essentially the same task.

## Testing Instructions
* Rebundle and build
* Place apiary markers in PA, IL and non-supported states/areas
* Verify that data is returned (or nodata) for the appropriate selection and that it renders correctly on the card.
* Test that both 3km and 5km options are working

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,3 +294,23 @@ Now any changes made to the front-end within `src/icp/apps/beekeepers` will be r
 **Note**: After running `start`, `build` must be run again to enable the front-end. This is because the template reads the `apps/beekeepers/.webpack.stats.json` file to decide which JavaScript files to include, which are generated and stored in the static files directory with `build`. However, `start` serves those files from its own port, which is no longer active when it is shut down, so we must run `build` again to generate the static files.
 
 The [`beekeepers.sh`](./scripts/beekeepers.sh) script just SSH's in to the `app` VM and runs [`yarn.sh`](./src/icp/apps/beekeepers/yarn.sh), so it can take any `yarn` parameters.
+
+## Apiary Score raster layers
+The raster data layers read by the application are hosted in the icp-bees AWS account and
+are used by dev, staging and production (there is only read-only access). The data layers
+are provided in pairs, one each at 3km and 5km focal operations. Additionally, each layer
+is provided per state and the file naming scheme represents this as:
+
+`[STATE_ABBR]_[INDICATOR]_[FORAGE_RADIUS].tif`
+
+This results in files named like `PA_pesticide_3km.tif`.
+
+In order to support multiple states that come as discrete files, a VRT per layer/radius pair
+has been created using the following steps:
+
+```bash
+gdalbuildvrt pesticide_3km.vrt PA_pesticide_3km.tif IL_pesticide_3km.tif
+```
+
+Once a VRT has been created for each layer/radius, both VRT and tifs are uploaded
+to the data bucket under a folder indicating 3km or 5km.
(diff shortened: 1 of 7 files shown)


src/icp/apps/beekeepers/views.py
```diff
@@ -17,6 +17,8 @@
 
 
 DATA_BUCKET = os.environ['AWS_BEEKEEPERS_DATA_BUCKET']
+# TODO: Remove before merging.  Temporary for PR Testing
+DATA_BUCKET = 'temp-beekeepers-azavea-us-east-1'
```
mmcfarland: Note to self: remove before merging.  Copy data to conventional bucket location as well.

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
```diff
@@ -50,9 +50,9 @@ const ApiaryCard = ({ apiary, forageRange }) => {
                 <ScoresLabel
                     indicator="forage"
```
fungjj92: The `indicator` passed to `<ScoresLabel />` should be updated to "floral"

mmcfarland: That prop seems to affect the label used on the card.  I had left this comment in the commit message:
> ...but the label "Forage" was left in the UI intentionally for the time being.

as I'm not sure what's best language to use for the beeks themselves, but I suspect it's forage.

README.md
```diff
@@ -294,3 +294,23 @@ Now any changes made to the front-end within `src/icp/apps/beekeepers` will be r
...
 
 The [`beekeepers.sh`](./scripts/beekeepers.sh) script just SSH's in to the `app` VM and runs [`yarn.sh`](./src/icp/apps/beekeepers/yarn.sh), so it can take any `yarn` parameters.
+
+## Apiary Score raster layers
+The raster data layers read by the application are hosted in the icp-bees AWS account and
+are used by dev, staging and production (there is only read-only access). The data layers
+are provided in pairs, one each at 3km and 5km focal operations. Additionally, each layer
+is provided per state and the file naming scheme represents this as:
+
+`[STATE_ABBR]_[LAYERNAME]_[FORAGE_RADIUS].tif`
```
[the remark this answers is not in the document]
mmcfarland: Cool, updated for consistency.

mmcfarland:
Pushed some fixups and also a script that auto-generates the VRTs based on the rasters we have now.  Additionally, we just received rasters for Indiana so I included them here.  The staging data rasters have been deleted and replaced by the new versions + the VRTs.

screenshot from 2018-12-12 18 32 37

src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
fungjj92: Oops! Missed the commit message.

PR #387, opened 2018-12-18
Make map interaction elegant

fungjj92:
## Overview

Previously, any click to any part of the map registered as both adding a marker/apiary and whatever else the action was _supposed_ to be. ie. a double click zoomed AND added a marker. This PR decouples single clicks adding a marker from double click zooms and clicking on map controls.

Connects #320 

### Demo

no-control-click

dblclick

### Notes

As noted in a commit comment, I lifted code from https://github.com/azavea/Leaflet.favorDoubleClick as it was more efficient than updating the library on the bees budget.

Random, I moved the survey component to the components folder because it makes more sense.

## Testing Instructions

`./scripts/beekeepers.sh start` and ensure using the map is as you expect


src/icp/apps/beekeepers/js/src/components/Map.jsx
```diff
@@ -60,29 +65,46 @@ class Map extends Component {
     onClickAddMarker(event) {
         const { forageRange, apiaries, dispatch } = this.props;
 
-        // Traffic cop, prevent simultaneous, clobbered updates to the state
-        // by only allowing 1 new apiary at a time
-        if (apiaries.find(a => !!a.fetching)) {
+        // Ignore clicks on a map control
+        if (event.originalEvent.target.className.includes('control')) {
```
kellyi: General question about this change:

It appears that this is the `onClick` handler for the React Leaflet map. Is there a different interaction for single clicks? Or is the map only supposed to respond to double clicks?

I believe you may be able to get the same functionality without these extra checks by replacing `onClick=` with `onDoubleClick=` or `onDblClick=` or similar in the map's props -- assuming there's no other `onClick` listener there to intercept.

I believe you'd also have to set `doubleClickZoom={false}`, since that's the usual behavior for responding to double clicks.

fungjj92: The map accepts single clicks (add marker), double clicks (map zoom) and clicks to interact with controllers

kellyi: Cool, that makes sense -- I misunderstood what interaction was supposed to be adding apiaries.

fungjj92: All good ^_^

fungjj92:
Just pushed up a fix to check for lat/lng matching in 73f2292

rajadain:
I'll take another look shortly.

rajadain:
I'm still seeing this behavior:



When I hover on an existing marker, the cursor changes to select, indicating the behavior is different. But clicking the marker drops another behind it, which is confusing. Can that be addressed?

fungjj92:
I had thought you meant clicking the same location erroneously creating another apiary, which was an issue. Since clicking a marker doesn't do anything, I'll update the CSS so the cursor doesn't change.

This wireframe has the cursor as the pointer, so I'll default the map cursor to always be that.

screen shot 2018-12-19 at 1 43 00 pm

fungjj92:
Merging ~

Issue #396, opened 2018-12-26
Add inset map to survey cards


fungjj92:
![screen shot 2018-12-26 at 10 30 44 am](https://user-images.githubusercontent.com/10568752/50450447-22caaf00-08fc-11e9-828d-668a91947838.png)

The inset map doesn't add that much value, and will take a notable amount of effort.

Connects #337

Issue #393, opened 2018-12-21
Styling pass list


fungjj92:
A place to accumulate styling issues for the next styling pass.

- [ ] Adding new apiary cards when the sidebar has vertical scroll adds a page scroll for a second
![flashscroll](https://user-images.githubusercontent.com/10568752/50348554-15ca4c00-0507-11e9-8d33-7e3500ab5b32.gif)

- [ ] What marker color should starred & surveyed apiaries get?

- [x] ~~Markers need better styling~~

- [x] Need to add real icons for the check and circle on survey cards, and the user dropdown caret

- [x] the survey view in general

- [x] Survey Detail Dialog in particular

- [x] Long apiary names overflow

- [x] All forms could use some design work, especially aligning labels and checkboxes.

- [x] Sidebar doesn't scroll in Firefox (the whole page scrolls instead):
    ![2019-01-10 11 59 50](https://user-images.githubusercontent.com/1430060/50984291-88f42c00-14cf-11e9-9b7f-47961425f766.gif)

- [x]  Gray highlight around survey links in Chrome 65 and Firefox. See:
![image](https://user-images.githubusercontent.com/10568752/51194973-24afde80-18ba-11e9-8f03-8f88371e3e98.png)

- [ ] Add styling to the "Retry" fetch data request button:
<img width="408" alt="screen shot 2019-01-15 at 5 40 06 pm" src="https://user-images.githubusercontent.com/10568752/51494953-9a1f2180-1d88-11e9-90e4-a18405e4c7c3.png">

rajadain:
Hey @jfrankl it would be great if you could take a look at these issues above. Let us know if we can help / pair / provide context or assistance. Thanks.

jfrankl:
@rajadain sure thing. I'm planning to work on styling cleanup tomorrow and will reply to these issues then.

jfrankl:
For the "Retry" fetch data error, it's possible to have multiple of those accumulate on the screen. It's also not possible to remove them, aside from "Retry" and hoping that works.

![image](https://user-images.githubusercontent.com/1809908/51952565-f9bfa180-2406-11e9-8d40-874c2c3dad13.png)

For now, let's add an "X" button to close these elements.

Issue #337, opened 2018-11-13
Make a Component for Apiary's Survey Listing


fungjj92:
![image](https://user-images.githubusercontent.com/10568752/48436907-f133c500-e74d-11e8-940a-2cf18f5eb8a0.png)

- Make a request the apiary list endpoint on the survey page
- Create a component that takes an apiary and can generate the two types of survey apiary cards (list each survey `monthyear` with status, or not participating)
- Sort the apiaries under the 3 header types

Requires #330, #334 , #336, list apiaries endpoint

PR #397, opened 2018-12-26
Add survey cards and sort them correctly

fungjj92:
## Overview

Builds out the survey feature by sorting and displaying the survey cards. The only button or link that works is the "add to study" blue button. The rest go to root, for now.

There's basic styling, but there's much left to be desired. I've added some items to #393. I also didn't implement the map inset for time savings, I think the feature should be an enhancement instead #396 

Some assumptions were made about sorting and displaying data:
- Only month years between when the apiary was created and now will be avail for survey data collection
- Month years available to survey includes the current month (i.e. December-2018 as I write this PR)
- For April/Nov there is no specific check implemented for if a subsurvey for that month is Monthly or specialty April/Nov survey for that month. I figured that could be handled later..

Connects #337

### Demo

screen shot 2018-12-26 at 12 23 37 pm

## Testing Instructions

Ensure your database is up to date on migrations (none in this PR, but you'll need to play with the data).
- Login
- Create some apiaries and toggle some of them to be "surveyed"
- Select an apiary that you will work with. In the django shell:
  - change that apiary's `created_at` date to some months ago. for example:
```
from datetime import datetime
from apps.beekeepers.models import Apiary
a = Apiary.objects.first()
a.created_at = datetime(year=2017, month=1, day=12)
a.name = "Jenny's Apiary"
a.save()
```
I also found it useful to change the name to something identifiable
- In the Django admin interface (easiest way) add a few surveys with subsurveys to an apiary in the time since said apiary was created (i.e. Jan 2017, as above).
- Visit http://localhost:8000/survey/?beekeepers and see the cards in different states.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
@@ -3,15 +3,19 @@ import { connect } from 'react-redux';
 
 import { Apiary } from '../propTypes';
 
+import { listMonthYearsSinceCreation } from '../utils';
+
 const SurveyCard = ({ apiary }) => {
     const {
         name,
         // surveyed,
         // monthyear,
     } = apiary;
-    
-    const surveyedMonthYears = apiary.surveys.map(s => (
-        <a href="/">{s.month_year}</a>
+
+    const monthYears = listMonthYearsSinceCreation(apiary);
+
+    const surveyedMonthYears = monthYears.map(m => (
+        <a href="/">{m}</a>
     ));
 
     return (
--- a/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
@@ -33,15 +33,16 @@ const SurveyCard = ({ apiary, dispatch, completed }) => {
         );
     } else {
         const monthYears = listMonthYearsSinceCreation(apiary);
+        const lastFourMonthYears = monthYears.slice(0, 4);
         if (completed) {
-            cardBody = monthYears.map(m => (
+            cardBody = lastFourMonthYears.map(m => (
                 <div className="listing">
                     <div className="listing__icon--completed">✓</div>
                     <a className="listing__monthYear" href="/">{m}</a>
                 </div>
             ));
         } else {
-            cardBody = monthYears.map(m => (
+            cardBody = lastFourMonthYears.map(m => (
                 <div className="listing">
                     <div className="listing__icon">◯</div>
                     <a className="listing__monthYear" href="/">{m}</a>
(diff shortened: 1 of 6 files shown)


src/icp/apps/beekeepers/js/src/components/SurveyView.jsx
```diff
@@ -4,14 +4,28 @@ import { arrayOf } from 'prop-types';
...
+
+    surveyedApiaries.forEach((a) => {
+        const monthYearsSinceCreation = listMonthYearsSinceCreation(a);
+        const apiarySurveyDates = a.surveys.map((s) => {
+            const monthName = monthToText(Number(s.month_year.substring(0, 2)) - 1);
+            const year = s.month_year.substring(2, 6);
+            return `${monthName}-${year}`;
+        });
+        if (monthYearsSinceCreation.every(date => apiarySurveyDates.find(d => d === date))) {
+            return surveyCards.complete.push(<SurveyCard apiary={a} />);
```
mmcfarland: The `return` here and below are a little confusing since the `forEach` doesn't assign a value.  Since it's just being used to break the current step of the iteration, consider removing them and using an `if/else` instead of the two returns while assigning to the array.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
```diff
@@ -0,0 +1,94 @@
...
+                    onClick={onSurvey}
+                >
+                    Add to study
+                </button>
+            </>
+        );
+    } else {
+        const monthYears = listMonthYearsSinceCreation(apiary);
+        const lastFourMonthYears = monthYears.slice(0, 4);
+        if (completed) {
```
mmcfarland: This looks like it's conflating "completed" in the sense that the whole thing is completed or not, rather than it could have some surveys completed and some not.  For example, if you change the Apiary `created_at` field to a year ago, then manually add a survey for Dec-2018, it still shows up with "incomplete" circle icon for December 2018, where it should actually be a check and the other items should remain incomplete.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
```diff
@@ -0,0 +1,94 @@
...
+                    className="button--long"
+                    onClick={onSurvey}
+                >
+                    Add to study
+                </button>
+            </>
+        );
+    } else {
+        const monthYears = listMonthYearsSinceCreation(apiary);
+        const lastFourMonthYears = monthYears.slice(0, 4);
```
mmcfarland: I'm not sure where the last 4 concept came from.  Did this come out of the design meeting?  My most recent understanding was that there were two columns and we could show the last years worth.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
fungjj92: doh!

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
mmcfarland: That's fine with me, just confirming that there'd be a way to get to a larger list.

fungjj92:
Thanks for taking a look! All your comments were spot on. I've made the related changes, and will break out a card for adding in the lite/pro user check #398

Issue #383, opened 2018-12-12
Add Survey Form for User after Sign Up


rajadain:
After a user signs up, they need to fill out a survey to participate in the program. They need to select a level of contribution (Light or Pro), and fill in details as specified in the User tab of this [spreadsheet](https://docs.google.com/spreadsheets/d/1lkXvBpeyg-lhzH48CdBe_C-HuepJBaXUM2904umd7PI). All users _must_ fill this out.

rajadain:
Follow this style:

![image](https://user-images.githubusercontent.com/1430060/50245913-ccacb780-03a1-11e9-959b-895e0a625b50.png)

https://share.goabstract.com/5684a5ec-fec9-41eb-9c16-6d5b9d104c7f

PR #400, opened 2018-12-28
User survey form

fungjj92:
## Overview

After a user logs in the for the first time to the beekeepers app, they need to fill out a user survey that details some basic information. The user shouldn't be able to use the app or exit without filling out this survey, though the fields are mostly optional.

The only field that isn't optional (from the UI's perspective) is `year_began`. I did this only because the django model expects a positive int. In the UI, the text input can't be given a default value, even null, `''`' or `NaN`. It was easier to just require this simple piece of info.

Connects #383

### Demo

screen shot 2018-12-28 at 4 04 21 pm

### Notes

Forms are crazy!! Took a long time to get just right enough... for all who do a survey form, we should pair because i definitely learned a thing or two here that would save time in the future (but not worth writing down perse)

Most importantly though was loosening up the Django model to be form-ready: basically, fields accepting blanks in most cases. See the commit message or https://stackoverflow.com/questions/8609192/differentiate-null-true-blank-true-in-django

If the user survey fails, the error handling is a generic message -- which isn't helpful. Unlike logging in, however, the error messages that I've seen come back were not very helpful in context or tough to parse/not worth it.

## Testing Instructions
`./scripts/manage migrate`
Assuming you have a user, log in to beekeepers. 
- Logging in when you don't have the user survey yet should prompt the user survey modal
The survey should submit just fine without entering more than the prompted required field(s).

- Logging into Bees 1 should work normally
- Logging in when you already have a user survey shouldn't prompt the user survey modal

FYI the new endpoint is at `/beekeepers/user_survey/`


src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,475 @@
...
+        const income = arrayToSemicolonDelimitedString(
+            [income_none, income_rent, income_sell_honey, income_sell_nucs, income_sell_queens],
+        );
+        const prepended_equipment_other = equipment_other.length ? `OTHER-${equipment_other}` : '';
+        const equipment = arrayToSemicolonDelimitedString(
+            [equipment_8_frame, equipment_10_frame, equipment_top_bar, prepended_equipment_other],
+        );
+        const form = Object.assign({}, this.state);
+        form.income = income;
+        form.equipment = equipment;
```
kellyi: Just a heads up that you can add these keys in a single pass by doing:

```
const form = Object.assign({}, this.state, {
    income,
    equipment,
});
```

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,475 @@
...
+            equipment_10_frame: '',
+            equipment_top_bar: '',
+            equipment_other: '',
+        };
+        this.handleChange = this.handleChange.bind(this);
+        this.handleSubmit = this.handleSubmit.bind(this);
+    }
+
+    handleChange(event) {
+        const target = event.currentTarget;
```
kellyi: In practice I'm not sure there's much difference but I think we've typically used `event.target` rather than `event.currentTarget` for this. A quick search indicated that the difference is that `target` is the element on which an event occurred while `currentTarget` is the element on which the event handler has been bound. Seems like those will be the same thing here, so I don't know whether or how much the difference matters.

Either way, potentially could save a few lines here by writing the signature as:

```
handleChange({ currentTarget: { value, type, checked, name } }) {
```

rajadain:
Build is failing because of lint issues in `user/views.py`:

```
$ ./scripts/check.sh
+ vagrant ssh app -c '(flake8 /vagrant/scripts/colors && flake8 /opt/app/apps --exclude migrations,node_modules) || echo flake8 check failed'
/opt/app/apps/user/views.py:36:1: W293 blank line contains whitespace
/opt/app/apps/user/views.py:42:80: E501 line too long (87 > 79 characters)
/opt/app/apps/user/views.py:72:80: E501 line too long (83 > 79 characters)
flake8 check failed
Connection to 127.0.0.1 closed.
```

src/icp/apps/beekeepers/models.py
```diff
@@ -368,10 +365,11 @@ class UserSurvey(models.Model):
         ),
         help_text='Do you prefer email or phone?')
     year_began = models.PositiveIntegerField(
-        null=False,
+        null=True,
```
rajadain: Since this is required in the UI, we should keep it non-NULL in the back-end.

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,475 @@
...
+                closeOnEscape={false}
+                closeOnDocumentClick={false}
+                modal
+            >
+                <div className="authModal">
+                    <div className="authModal__header">
+                        <div>User survey</div>
+                    </div>
+                    <div className="authModal__content">
+                        <div className="title">Fill our your user survey</div>
```
rajadain: Typo: "Fill ~~our~~ out your user survey"

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,475 @@
...
+                                <br />
+                                <input
+                                    type="checkbox"
+                                    className="form__control"
+                                    name="income_sell_nucs"
+                                    checked={income_sell_nucs}
+                                    onChange={this.handleChange}
+                                    value="SELL_NUCS_PACKAGES"
+                                />
+                                Sell nucs packages
```
rajadain: Can we switch the order of the checkbox here, or make it on the same line? Currently the checkbox is sometimes on the line after, and sometimes on the line before, and it would be good to be consistent:

image

fungjj92:
Changes pushed up, thanks for the looks! Still getting used to object destructuring.

rajadain:
Taking another look now.

rajadain:
For some reason, the checkbox values aren't being recorded correctly. For example, in this case, a number of checkboxes are checked:

screen shot 2019-01-02 at 5 07 55 pm

But what is sent is either `false` or `"false"`:

image

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    handleChange({
+        currentTarget: {
+            value, type, checked, name,
+        },
+    }) {
+        let finalValue = value;
+        if (type === 'checkbox') {
+            if (value === 'on') {
```
rajadain: I think this is the problem. Not sure what the `on` check is about. This seems to fix the issue:

```diff
diff --git a/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx b/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
index f701cec4..37acc6c2 100644
--- a/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
@@ -48,14 +48,8 @@ class UserSurveyModal extends Component {
             value, type, checked, name,
         },
     }) {
-        let finalValue = value;
-        if (type === 'checkbox') {
-            if (value === 'on') {
-                finalValue = checked;
-            } else {
-                finalValue = checked ? value : '';
-            }
-        }
+        const finalValue = type === 'checkbox' ? checked : value;
+
         this.setState({ [name]: finalValue });
     }
 
```

rajadain: It was showing up as `"false"` in my testing, even when `checked` was `true`

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -0,0 +1,494 @@
...
+                                </label>
+                                <label htmlFor="income_rent">Rent colonies for pollination</label>
+                                <input
+                                    type="checkbox"
+                                    className="form__control"
+                                    id="income"
+                                    name="income_rent"
+                                    checked={income_rent}
+                                    onChange={this.handleChange}
+                                    value="RENT_COLONIES_FOR_POLLINATION"
```
rajadain: Why does this checkbox have this value?

rajadain: Oh, so that they're compiled together upon submission. Hmm. Not sure if we should be overloading `value` like this, as it can change when the user checks or unchecks the box.

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
fungjj92: what browser are you using, btw? I'm not seeing the errors you are.

rajadain: Chrome 71:

image

fungjj92:
It took a while for me to see, but the issue was that checkbox inputs should use the `checked` prop not the `value` prop, unlike other input types, see 982e915. Good catch! I am also using Chrome 71 FYI.

screen shot 2019-01-03 at 12 58 09 pm

screen shot 2019-01-03 at 1 01 26 pm

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
fungjj92: `value` doesn't change for checkboxes. Only the `checked` prop does (true/false)

rajadain:
Taking another look.

fungjj92:
Thanks for reviews! Many lessons learned for the next survey cards 👊

Issue #335, opened 2018-11-13
Implement signing out


fungjj92:
If a user is logged in, the dropdown of their name should have a "Sign out" button. Includes adding the dropdown.

No wireframe exists, however this is fairly intuitive without.

![image](https://user-images.githubusercontent.com/10568752/48428420-f9353a00-e738-11e8-8fdf-de5fa1454cd2.png)

Requires #333

PR #401, opened 2018-12-31
Implement Logging Out

rajadain:
## Overview

Implements logging out.

Connects #335 

## Testing Instructions

* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and log in. Ensure you see a ▾ icon next to the username.
* Hover on the username. Ensure you see a dropdown with a "Log Out" button.
* Ensure hovering on the username, or the Log Out button, keeps the dropdown open and it doesn't accidentally close.
* Click the "Log Out" button. Ensure you are logged out and the app resets to its initial state.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -167,6 +167,17 @@ export function login(form) {
     };
 }
 
+export function logout() {
+    return dispatch => csrfRequest.get('/user/logout').then(() => {
+        dispatch(setAuthState({
+            username: '',
+            authError: '',
+            userId: null,
+        }));
+        dispatch(setApiaryList([]));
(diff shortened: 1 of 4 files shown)


fungjj92:
Add the carrot icon to this list#393

rajadain:
Good point about accessibility. Generally any app that is primarily map driven is going to be a challenge for the visually impaired, so we may have bigger problems, but it's a good habit to make things accessible by default, which we should always try to do. 5ae9b7e includes a hidden button that will be visible to screen readers only, using [recommended styling from 18F](https://accessibility.18f.gov/hidden-content/).

fungjj92:
Great ! Tested on ipad mode on chrome.

rajadain:
Thanks for taking a look!

Issue #402, opened 2018-12-31
Implement "Forgot Password" and "Resend Activation Link" Workflows


rajadain:
These are alluded to in the Beekeepers UI but not properly implemented. The endpoints are already available, just need to wire up the UI correctly.

Issue #328, opened 2018-11-13
Save unsaved apiaries upon sign up


fungjj92:
If an unauthenticated user has persisted apiaries in their localstorage and meanwhile signs up for and validates an account, the cached apiaries should now be saved to the database to their new user.

Requires #326, #327, apiary endpoints and model existing in Django

Issue #326, opened 2018-11-13
Implement Sign Up workflow


fungjj92:
If not already added, sign up (and login) will use the two-step activation workflow under [Django Registration](https://django-registration.readthedocs.io/en/3.0/) where the user will receive email activation links to activate the account.

Assuming #290 is implemented, users already signed up with the Pollinator app trying to sign up with the will have to be handled a little differently. The user should be able to use their same credentials between apps. When the user submits the form, if their user exists in the user tab, we ignore the form inputs and send an email verification link with copy explaining to use their Pollinator login information to log in. After the link is clicked the the beekeepers access column for the existing user becomes set to `True` and the same user and access both apps. 

Hook up to the modal implemented in #325 

![image](https://user-images.githubusercontent.com/10568752/48427364-93e04980-e736-11e8-8bd9-d3b8c61516d6.png)

Requires #325

PR #403, opened 2018-12-31
Implement Sign Up workflow

rajadain:
## Overview

Expands the Sign Up dialog to have four entry fields for email, username, and two password fields. All fields are required, and HTML5 validation is used to ensure they are filled before sending it to the server.

Once complete, the data is sent to the `/user/sign_up?beekeepers` endpoint as form data, which returns validation errors or a success response. The first validation error is shown at the top of the dialog, using similar styling as in the Log In dialog. The query string is used to classify the new users as having signed up from the Beekeepers app,  as opposed to Pollination app.

If the sign up is successful, the Sign Up dialog is closed and the Log In dialog is opened, with a message prompting the user to click the activation link in their email. Once they have activated their account, they can log in with the specified credentials. Upon first log in, any work they have done so far (in terms of apiaries created) are saved to the database.

Connects #326 
Connects #328 

### Notes

* The issue mentions something about enabling Pollination users for Beekeepers. I don't think there's such a separation anymore, so that can be disregarded.

* I noticed that when a user is created via this sign up dialog, in the `user_userprofile` table they are still listed as coming from the Pollination app. This is likely because the check for HTTP_REFERRER doesn't correctly capture the querystring parameter or something:

    https://github.com/project-icp/bee-pollinator-app/blob/3a82e6753d7dc572714e27977b839cd89d936c52/src/icp/apps/user/views.py#L127-L133

    ```sql
    SELECT username, origin_app
    FROM auth_user u INNER JOIN user_userprofile p ON u.id = p.user_id
    WHERE u.username = 'test4';
    
     username | origin_app  
    ----------+-------------
     test4    | Pollination
    (1 row)
    ```

    When this changes to looking at the domain name, I think this'll fix itself.

* "Forgot Password" and "Resend Activation Link" workflows are alluded to, but not implemented herein. That is deferred until later in #402.

* The activation link takes the user to the Pollination Mapper app. There's very little content on the page, so I think it's alright for now.

## Testing Instructions

* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and ensure you are logged out
* Open the sign up dialog and sign up. Try not filling out all fields, making the passwords mismatch, an invalid email, and using a username that already exists. Ensure you get proper error messages for each.
* Run `./scripts/debugserver.sh`
* Fill in valid details in the Sign Up modal. Ensure you are redirected to the Log In dialog.
* Try to log in. Ensure it asks to activate the account first.
* In your `debugserver` output, find the activation link. Open it.
* Try logging in again. Ensure it works correctly.
* Log out, and create some apiaries. Log in again, and ensure those apiaries are still there.
* Refresh the page while logged in and ensure the apiaries are still there.
* Log out, and create some apiaries. Log in again, and ensure you see the apiaries already saved in the account, not the new ones you created while logged out.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/App.jsx
+++ b/src/icp/apps/beekeepers/js/src/App.jsx
@@ -17,7 +17,7 @@ import ParticipateModal from './components/ParticipateModal';
 import UserSurveyModal from './components/UserSurveyModal';
 
 import { UserSurvey } from './propTypes';
-import { login, fetchUserApiaries, openUserSurveyModal } from './actions';
+import { login, saveAndFetchApiaries, openUserSurveyModal } from './actions';
 
 class App extends React.Component {
     componentDidMount() {
@@ -30,9 +30,9 @@ class App extends React.Component {
         const { dispatch, userId, userSurvey } = this.props;
 
         if (userId && userId !== prevProps.userId) {
-            if (userSurvey) {
-                dispatch(fetchUserApiaries());
-            } else {
+            dispatch(saveAndFetchApiaries());
+
+            if (!userSurvey) {
                 dispatch(openUserSurveyModal());
             }
         }
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -333,3 +333,70 @@ export function deleteApiary(apiary) {
         }
     };
 }
+
+export function saveAndFetchApiaries() {
+    return (dispatch, getState) => {
+        const {
+            main: {
+                apiaries,
+            },
+            auth: {
+                userId,
+            },
+        } = getState();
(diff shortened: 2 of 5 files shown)


fungjj92:
This works well. The activation page has copy that doesn't apply to beekeepers however, which is that logging in saves ongoing work. I think we had pushed this workflow to enhancement or removed it altogether.

rajadain:
I fixed the app sourcing issue in 7b14b5b, and added a commit for #328 in 3264e22. Could you take another look at this?

rajadain:
Rebased over `develop` to resolve merge conflicts.

fungjj92:
Works well, tested after rebase !

rajadain:
Thanks for taking a look! Merging now.

Issue #404, opened 2019-01-03
Fix Failing Builds


rajadain:
In https://github.com/project-icp/bee-pollinator-app/commit/462882e1248198c0f630a877bd0e7b75bd8668d9, I changed some files to be symlinks since they’re used in both apps. This works correctly on my machine when I do `./scipts/bundle.sh --vendor --tests`, and when I do `vagrant reload app --provision`. However, this very command and step is failing on CivicCI01: http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-app-and-worker/107/console.

PR #405, opened 2019-01-03
Fix Build Failures by Copying Symlinks as Target Files

rajadain:
## Overview

A few files were converted to be symlinks in https://github.com/project-icp/bee-pollinator-app/commit/462882e1248198c0f630a877bd0e7b75bd8668d9. While this works for Vagrant, this broke for AMI builds. By adding `copy_links=yes` according to the [synchronize module](https://docs.ansible.com/ansible/latest/modules/synchronize_module.html), we ensure they are copied as the targets and not symlinks and work correctly.

Connects #404 

### Demo

http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-app-and-worker/109/console
http://civicci01.internal.azavea.com/view/bees/job/bee-pollinator-staging-deployment/108/

The map legend (which relies on the symlinked files) is working correctly on staging:

image

### DIFF
--- a/deployment/ansible/roles/bee-pollinator.app/tasks/app.yml
+++ b/deployment/ansible/roles/bee-pollinator.app/tasks/app.yml
@@ -17,6 +17,7 @@
 
 - name: Synchronize Django application into app_home
   synchronize: archive=no
+               copy_links=yes
                checksum=yes
                compress=no
                recursive=yes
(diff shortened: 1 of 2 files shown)


rajadain:
Thanks for pointing me in the right direction and reviewing!

Issue #398, opened 2018-12-26
List surveys based on Lite/Pro user status


fungjj92:
A bit of work broken off #397 as it's scope is already quite large for a card.

As of #397, the survey cards are rendered and subsurveys listed universally as: monthly for all months between the apiary created date and now, inclusive. This is a first step but doesn't encompass the unique survey requirements of lite and pro.

It would be nicer to handle this after we implement the user survey workflow, but not necessary as the user survey model exists.

Issue #378, opened 2018-12-12
Make a Survey Listing Detail Modal


rajadain:
The survey listing card shows only the 4 most recent surveys. Clicking on "View full history" opens a modal which shows all the surveys related to this apiary. Implement the detail modal:

![image](https://user-images.githubusercontent.com/1430060/49892628-7080eb00-fe17-11e8-9e48-d94bc08745e9.png)

![image](https://user-images.githubusercontent.com/1430060/49892636-77a7f900-fe17-11e8-9b9a-02b4822bc604.png)

rajadain:
Shares points with #398

PR #406, opened 2019-01-03
List Surveys Based on User Contribution Level

rajadain:
## Overview

Updates the Survey Listing for each apiary to list November, April, and Monthly surveys separately, with explicit marking. Pro users see all three kinds of listings, and Light users see only November and April listings. We begin counting from November 2018 for November and April surveys, and from the date of the apiary's creation for the Monthly surveys.

We still show the four most recent survey listings in a survey card, with the rest hidden behind a "View full history" button. Clicking that button opens a popup that lists all the surveys for that apiary.

Connects #398 
Connects #378 

### Demo

For Pro Users:

image

For Light Users:

image

### Notes

The Survey Detail dialog needs styling. I've made a note in #393.

## Testing Instructions

* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and log in as a Pro user.
* Make some apiaries. Opt some in to the survey. Switch to the survey tab. Make sure you see an entry for November 2018, and a Monthly entry for each month since the apiary's creation.
* Go to [:5433](http://localhost:5433) and update one of the apiaries to have an older creation date. Refresh the survey page and ensure you see more entries for Monthly surveys
* If any apiary has more than 4 survey listings, ensure you see the "View full history" button, and clicking it opens the detail dialog, which has the same list but now complete, not just limited to the last 4.
* Log out. Log in as a Light user. Make some apiaries and opt them in to the survey. Switch to the survey tab. Ensure you only see entries for November 2018, and none for Monthly surveys.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
@@ -1,12 +1,14 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { func } from 'prop-types';
-import { Apiary } from '../propTypes';
+import { arrayOf, func } from 'prop-types';
+import { Apiary, Survey } from '../propTypes';
 
-import { listMonthYearsSinceCreation, monthToText } from '../utils';
+import {
+    monthNames2,
+} from '../utils';
 import { setApiarySurvey } from '../actions';
 
-const SurveyCard = ({ apiary, dispatch }) => {
+const SurveyCard = ({ apiary, dispatch, surveys }) => {
     const {
         name,
         surveyed,
@@ -32,33 +34,26 @@ const SurveyCard = ({ apiary, dispatch }) => {
             </>
         );
     } else {
-        const monthYears = listMonthYearsSinceCreation(apiary);
-        const lastFourMonthYears = monthYears.slice(0, 4);
-        const apiarySurveyDates = apiary.surveys.map((s) => {
-            const monthName = monthToText(Number(s.month_year.substring(0, 2)) - 1);
-            const year = s.month_year.substring(2, 6);
-            return `${monthName}-${year}`;
-        });
-        cardBody = lastFourMonthYears.map((m) => {
-            const i = apiarySurveyDates.findIndex(date => date === m);
-            if (i >= 0) {
-                return (
-                    <div className="listing" key={name + m}>
-                        <div className="listing__icon--completed">✓</div>
-                        <a className="listing__monthYear" href="/">{m}</a>
-                    </div>
-                );
-            }
-            return (
-                <div className="listing" key={name + m}>
-                    <div className="listing__icon">◯</div>
-                    <a className="listing__monthYear" href="/">{m}</a>
-                    <a className="listing__start" href="/">Start survey</a>
-                </div>
-            );
-        });
+        const shownSurveys = surveys.slice(0, 4);
 
-        if (monthYears.length > 4) {
+        cardBody = shownSurveys.map(({ month_year, survey_type, completed }) => (
+            <div className="listing" key={name + month_year + survey_type}>
+                <div className={`listing__icon${completed ? '--completed' : ''}`}>
+                    {completed ? '✓' : '◯'}
+                </div>
+                <a className="listing__monthYear" href="/">
+                    {monthNames2[month_year.slice(0, 2)]}
+                    {' '}
+                    {month_year.slice(-4)}
+                    {' '}
+                    (
+                    {survey_type}
+                    )
+                </a>
+            </div>
+        ));
+
+        if (surveys.length > 4) {
             cardFooter = (
                 <div className="surveyCard__footer">
                     <button type="button" className="button">View full history</button>
@@ -91,6 +86,7 @@ const SurveyCard = ({ apiary, dispatch }) => {
 SurveyCard.propTypes = {
     apiary: Apiary.isRequired,
     dispatch: func.isRequired,
+    surveys: arrayOf(Survey).isRequired,
 };
 
 export default connect()(SurveyCard);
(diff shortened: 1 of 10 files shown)


src/icp/apps/beekeepers/js/src/components/SurveyCardListing.jsx
```diff
@@ -0,0 +1,27 @@
...
+        <div className={`listing__icon${completed ? '--completed' : ''}`}>
+            {completed ? '✓' : '◯'}
+        </div>
+        <a className="listing__monthYear" href="/">
+            {monthNames[monthYear.slice(0, 2)]}
+            {' '}
+            {monthYear.slice(-4)}
+            {' '}
+            (
+            {type}
```
rajadain: Perhaps this type should only be shown for Pro users, and not Light users, because the latter will only see "November 2018", "April 2019", etc which is self explanatory.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -16,3 +16,10 @@ export const SORT_OPTIONS = [DEFAULT_SORT].concat(Object.values(INDICATORS));
 
 export const MAP_CENTER = [40.0, -76.079];
 export const MAP_ZOOM = 10;
+
+export const CONTRIBUTION_LEVEL_LIGHT = 'LIGHT';
+export const CONTRIBUTION_LEVEL_PRO = 'PRO';
+
+export const SURVEY_TYPE_NOVEMEBER = 'NOVEMBER';
```
fungjj92: typo: `SURVEY_TYPE_NOVEMBER`

src/icp/apps/beekeepers/js/src/propTypes.js
```diff
@@ -52,3 +52,9 @@ export const UserSurvey = shape({
     rear_queens: bool.isRequired,
     equipment: string,
 });
+
+export const SurveyItem = shape({
+    monthYear: string.isRequired,
+    type: string.isRequired,
+    completed: bool.isRequired,
```
fungjj92: The `SurveyItem` object will be workable for #408 and future surveys with the addition of a few more pieces of info, namely `apiaryId` and `surveyId`.

src/icp/apps/beekeepers/js/src/utils.js
```diff
@@ -139,3 +120,119 @@ export function toFormData(json) {
     Object.entries(json).forEach(([key, value]) => formData.append(key, value));
     return formData;
 }
+
+/**
+ * Given an Apiary, returns a list of November SurveyItem objects, like this:
```
fungjj92: comment clarification:
`Given an Apiary, returns a list of November Survey Item objects since 2018, like this:`

or something like that.  Do the same for `getAprilSurveyItems` and `getMonthlySurveyItems` where the latter references since apiary creation instead of X year.

src/icp/apps/beekeepers/js/src/components/SurveyCardDetails.jsx
```diff
@@ -0,0 +1,29 @@
...
+        </div>
+    );
+};
+
+SurveyCardDetails.propTypes = {
+    apiary: Apiary.isRequired,
+    surveyItems: arrayOf(SurveyItem).isRequired,
+};
+
+export default SurveyCardDetails;
```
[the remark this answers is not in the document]
rajadain: Since this is a sub-component of `SurveyCard`, I wanted to namespace it to indicate that. But I suppose the dialog could be opened from other places too, so I'll make that change.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
```diff
@@ -32,47 +35,38 @@ const SurveyCard = ({ apiary, dispatch }) => {
...
+            >
+                View full history
+            </button>
+        );
+        const cardDetails = surveys.length <= 4 ? null : (
+            <Popup
+                modal
+                trigger={cardDetailTrigger}
+            >
+                <ApiarySurveyListingModal apiary={apiary} surveys={surveys} />
```
fungjj92: We use `ApiarySurveyListingModal` inside a modal here, but the component itself doesn't need to be used that way. Can the component name drop `Modal`?

src/icp/apps/beekeepers/js/src/propTypes.js
```diff
@@ -52,3 +52,13 @@ export const UserSurvey = shape({
...
 });
+
+export const Survey = shape({
+    id: number,
+    month_year: string.isRequired,
+    num_colonies: number,
+    created_at: string,
+    survey_type: string.isRequired,
+    apiary: number.isRequired,
+    completed: bool.isRequired,
```
fungjj92: Thanks for updating this propType to `Survey`. Makes more sense and save work hooking up the survey forms! 

I'm noticing the `completed` field which is helpful & necessary to app logic, but not technically part of the `Survey` model. Gonna get it slide since I have no other ideas, but wanted to point it out.

src/icp/apps/beekeepers/js/src/utils.js
```diff
@@ -139,3 +120,155 @@ export function toFormData(json) {
...
 }
+
+/**
+ * Given a list of surveys, returns a function that takes a month_year and
+ * returns true the survey if found, otherwise null.
+ *
+ * @param {arrayOf(Survey)} surveys
+ */
+function makeSurveyMatcher(surveys) {
+    return (month_year) => {
```
fungjj92: nice ! i never think to return a function, but case in point it can be very clutch

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
rajadain: Good point. Fixed.

rajadain:
Thanks for reviewing again. Will merge when green.

Issue #379, opened 2018-12-12
Add Survey Form for April Style Survey


rajadain:
According to the April tab in the [spreadsheet](https://docs.google.com/spreadsheets/d/1lkXvBpeyg-lhzH48CdBe_C-HuepJBaXUM2904umd7PI).

rajadain:
Follow this style:

![image](https://user-images.githubusercontent.com/1430060/50245913-ccacb780-03a1-11e9-959b-895e0a625b50.png)

https://share.goabstract.com/5684a5ec-fec9-41eb-9c16-6d5b9d104c7f

rajadain:
Use checkboxes for multiple choice questions, option buttons for single choice questions, and text boxes for free text entry.

PR #408, opened 2019-01-04
Create and show April surveys via form

fungjj92:
## Overview

This PR adds a functional April survey form that displays completed and submits new April surveys.

TODO's:
- [x] set state values from completed survey data
- [x] disabled button styling
- [x] figure out how month_year will be shared to the component (rebase on #406)
- [x] close modal on successful post or show error

Connects #379 

### Demo

screen shot 2019-01-07 at 5 08 20 pm
screen shot 2019-01-08 at 9 41 27 am

Basic error handling:
screen shot 2019-01-07 at 5 09 57 pm

completed-change

### Notes

Things I'm thinking.

It would be awesome to have a custom component that all the monthly surveys inherit from, since their functionality will be the same I think. The only tricky thing is the forms are custom, and that puts a damper in reusability. --> Timebox explore in #380 

~As discussed on slack #react, I wasn't sure how to handle the requests to the survey endpoints. It doesn't make sense to me to invoke Redux since survey data is very much only used in the scope of the modal being open. This is what I have that is working, for now. Open to discussion!~ Decided on making requests from component.

## Testing Instructions

In `utils` change the `startYear` (line 192) in `getAprilSurveys` to 2018.
Visit http://localhost:8000/survey/?beekeepers and click one of the April (APRIL) surveys.
Fill out the survey if it doesn't exist, else see the results display.

Other survey listings should open an empty popup which will be subbed in with their survey form components when they exist.


src/icp/apps/beekeepers/js/src/components/AprilSurveyModal.jsx
```diff
@@ -0,0 +1,283 @@
...
+        const {
+            apiaryId,
+            surveyId,
+        } = this.props;
+        if (surveyId) {
+            getOrCreateSurveyRequest({
+                apiaryId,
+                surveyId,
+            }).then(({ data }) => this.setState({ completedSurvey: data }))
+                .catch(error => this.setState({ error }));
```
kellyi: Out of curiosity, what happens to these API request handlers if the survey component is closed by the user before the Promise has resolved or rejected?

src/icp/apps/beekeepers/js/src/actions.js
```diff
@@ -311,3 +311,11 @@ export function deleteApiary(apiary) {
         }
     };
 }
+
+export function getOrCreateSurveyRequest({ apiaryId, surveyId, form }) {
```
kellyi: Gave this a bit more thought and I think this should just be a Redux action which changes the app state, and in turn the component should retrieve the data from the state instead of calling this to make the request directly. A couple reasons why:

First, setting up a path for API requests which do not go through Redux introduces a new pattern to the app which is pretty different from existing API requests here and from those in other Civic Apps projects. This seems potentially confusing to us when we look back on the code in the future.

Likewise, since this specific function isn't currently related to Redux at present, it probably belongs in `utils` rather than `actions`, since `actions` customarily gathers functions which change the Redux store in some way. However, I also think it'd be odd to have it in `utils`.

Lastly, I think having a React component make API requests essentially doubles its complexity. Generally our React component implementations have been concerned only with rendering views in response to whatever props they are given -- especially since Redux actions sort of enable them not to have to handle the details of success and error states at all (except insofar a these are encoded in props which are typechecked with `prop-types` in development).

Moving the request into the component means that component will now also also have to handle the immediate success and error outcomes directly -- mapping success data into whatever shape and handling whatever kind of errors (including network connection errors and errors returned from the API response) -- along with its view logic.

fungjj92: Thoughtful points. I discussed with @rajadain yesterday.

Foremost in our minds, we're in 🐦 mode. 

You bring up valid points. I don't think departing from codified patterns is a bad thing, if it makes sense. In this case, saving survey data -- that doesn't need or want to be seen by any other part of the app -- adds some complexity. If Redux is a developer convenience, it feels more inconvenient than anything to shoe-in here. In this case, the component handling success/error is desirable and straightforward -- we munge data a little or show the error message as sent. If the modal is closed before the request returns -- that's functionally fine too. We don't want that data anyway, if the user closes the modal. Not great javascript, but functionally acceptable, even desirable.

I will take the advice to move the action to `utils` however, since I'm choosing _not_ to use redux.

rajadain:
Just merged #406. This will need to be rebased atop `develop` to resolve the merge conflicts.

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
```diff
@@ -0,0 +1,297 @@
...
+                            disabled={!!completedSurvey}
+                        />
+                    </div>
+                    {submitButton}
+                </form>
+            </>
+        );
+
+        // request to survey endpoint should happen when the popup opens (not before, as is now)
+        // potential solution? move popup to parent, close handling to parent too
```
rajadain: Do these comments still need to be here?

src/icp/apps/beekeepers/js/src/components/SurveyCardListing.jsx
```diff
@@ -1,24 +1,62 @@
...
+        month_year,
+        survey_type,
+        completed,
+    },
+}) => {
+    let formComponent;
+    switch (survey_type) {
+        case SURVEY_TYPE_APRIL:
+            formComponent = (
+                <AprilSurveyForm month_year={month_year} apiaryId={apiary} surveyId={id} />
```
rajadain: Would it be easier to give `AprilSurveyForm` (and the others that will follow) the entire `survey` as a prop rather than these pieces? That would be more flexible if in case we wish to use some other fields in the future.

src/icp/apps/beekeepers/js/src/components/SurveyCardListing.jsx
```diff
@@ -1,24 +1,62 @@
...
+            );
+            break;
+        case SURVEY_TYPE_NOVEMBER:
+            formComponent = <div />; // replace with nov form component
+            break;
+        default:
```
rajadain: Prefer adding an explicit case for `SURVEY_TYPE_MONTHLY` and a do this instead:

```js
default:
  throw new Error(`Unknown survey type: ${survey_type}`);
```

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
```diff
@@ -0,0 +1,297 @@
...
+
+function mapStateToProps(state) {
+    return state.main;
+}
+
+AprilSurveyForm.propTypes = {
+    apiaryId: number.isRequired,
+    surveyId: number,
+    month_year: string.isRequired,
+    dispatch: func.isRequired,
```
rajadain: Why do we need `dispatch` here?

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
```diff
@@ -0,0 +1,297 @@
...
+
+        const form = Object.assign({}, this.state, multipleChoiceState, { survey });
+
+        getOrCreateSurveyRequest({ apiaryId, form })
+            .then(({ data }) => {
+                this.setState({
+                    completedSurvey: data,
+                    error: '',
+                });
+                dispatch(fetchUserApiaries());
```
rajadain: Why fetch user apiaries here?

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
fungjj92: good catch

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
fungjj92: Necessary for the survey list to update the completed state of the surveys. See the gif in the demo section.

src/icp/apps/beekeepers/js/src/components/SurveyCardListing.jsx
fungjj92: about the same, but a little cleaner to feed in the whole survey. i can make that change

src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
rajadain: Answered by https://github.com/project-icp/bee-pollinator-app/pull/408#discussion_r246129322

fungjj92:
Thanks for reviewing! Will wait for Jenkins before merging.

Issue #407, opened 2019-01-04
Prepare 2017 CDL version for use


mmcfarland:
The 2017 CDL needs to be preprocessed so that it can be tiled and incorporated into the original PollinationMapper.

Source
https://www.nass.usda.gov/Research_and_Science/Cropland/Release/index.php

* [x] Acquire raw CDL raster
* [ ] Group & Reclassify according to https://github.com/project-icp/bee-pollinator-app/tree/develop/src/icp/pollinator/src/pollinator/reclass

The follow up work is detailed in #371, #373 and #377

rajadain:
Is the reclassification complete? Can we check the second box?

mmcfarland:
It is.  Reclassified raster can be found in the icp s3 bucket: `icp-cdl-data`

Issue #377, opened 2018-12-12
Update local CDL raster for crop yield model


mmcfarland:
The intermediary steps in #371 produce a tiff that can replace the grouped 2015 CDL version used locally in the Pollinator app.

This work should be coordinated with #373 so that the overlay and the model reflect the same data.

PR #410, opened 2019-01-04
Update the worker snapshot volume for 2017 CDL

mmcfarland:
## Overview

The CDL tif has been updated to the 2017 version and placed on the
snapshot identified by the id added to our packer template.

Connects #407 
Connects #377 

### Demo

I ran a scenario on the 2015 raster and another scenario on the 2017 version.  As you can see, one version had Cucurbits dropped between versions.

screenshot 2019-01-04 at 1 56 02 pm

### Notes

The 2017 CDL was reclassified according to the excellent notes in the readme.  The articaft, aside from being available on the snapshot volume, is in the publically available `icp-cdl-data`.  As the readme indicates, the 2017 prefixed version can be copied (sans prefix) to your host `/opt/icp-crop-data` directory in order to use/test it locally.  The other proportions remain largely the same.

## Testing Instructions

* Update your local CDL version, as described in the Notes above.
* Confirm you can still run the pollination model with reasonable results
* Confirm the snapshot ID is valid.

When this is merged, we can verify that it's being used on staging.

### DIFF
--- a/deployment/packer/template.js
+++ b/deployment/packer/template.js
@@ -41,7 +41,7 @@
             "ami_block_device_mappings": [
                 {
                     "device_name": "/dev/sdf",
-                    "snapshot_id": "snap-08429591cf7ae1fb7",
+                    "snapshot_id": "snap-0a40643aeb3029eb0",
                     "volume_type": "gp2",
                     "delete_on_termination": true
                 }


mmcfarland:
This was deployed to staging and you can see in this screenshot that the baseline scenario (taken on the 2015 version) differs from the current (2017) version - tree fruits (other) have been classified out.

screenshot from 2019-01-08 09 52 06

Issue #392, opened 2018-12-20
Apiary card spinner should apply to whole card


fungjj92:
The apiary card spinner has a few issues.

First, the spinner doesn't follow the card but rather appears positioned absolutely on the page. Scrolling down the sidebar results in detached cards from spinners, spinners floating in space.

Second, because the spinner only applies to the scores (not the apiary card header), a user can interact with the card (toggle a button or eventually change the name) but those changes disappear once the data comes back from the fetch because of out of sync updates to the state's apiary list. As a quick working solution, the spinner should replace the card's whole contents so that these interactions are not available to the user.

PR #413, opened 2019-01-09
Overlap spinner to entire apiary card

fungjj92:
## Overview

The spinners when apiary cards loaded covered only the scores, and were displayed absolute to the page, not the card. Therefore, when the card list grew long and scrollable, spinners would float in space over the list.

This card fixes the spinners and fixes them over the entire card, not just the scores -- this has the benefit of not showing the apiary name until it is successfully reverse geocoded.

I gave cards a minimum height of 80px, which is roughly their height already when loaded with info.

Connects #392 

### Demo

screen shot 2019-01-09 at 5 30 18 pm

## Testing Instructions

Click the map, add an apiary. Toggle the buffer. Add enough apiaries so the list scrolls. Spinners should act as you expect.


rajadain:
Taking a look now

fungjj92:
Thanks for review!

Issue #416, opened 2019-01-10
Improve Form Workflow


rajadain:
Currently when the user fills out their user form which appears on first login, it just disappears. This can be a bit jarring. ~~Add a prompt or message thanking the user, and introducing the next steps. Need to figure out exactly what should be shown. Talk to designers and PM before proceeding.~~

When a form is filled out, close the modal, then show a modal that says "Form Submitted" and some indication of success ✅. Optionally, the modal should close itself after a few seconds.

fungjj92:
I think this popup can apply for #402 as well, so I'm rolling them into one.

PR #418, opened 2019-01-11
Add forgot PW and resent activation link workflows

fungjj92:
## Overview

Take care of user management loose ends by adding forgot password and resend activation link workflows. I ended up doing in-modal messaging rather than connecting to the success modal in #416

Connects #402 

### Demo
error:
screen shot 2019-01-11 at 1 52 17 pm

success / next steps message:
screen shot 2019-01-11 at 2 04 40 pm

### Notes

I made the non-error message black like in the Bees1 app. Green/success felt wrong, it is more like instructions than a success message.

## Testing Instructions
Run both
`./scripts/beekeepers.sh`
`./scripts/debugserver.sh`

Sign out and click the login button. Play around with the forgot pw, resend activation link workflows. You'll need a signed up but inactive account to use the latter workflow.


fungjj92:
Thanks!

Issue #372, opened 2018-12-12
Make Hover Popups for Indicator Information


rajadain:
As shown in the mockup: 

![image](https://user-images.githubusercontent.com/1430060/49892224-6ca09900-fe16-11e8-9729-1cc974944fb4.png)

Hovering on a score should show additional information about the indicator. Blocked until we have text to add to the popup.

PR #420, opened 2019-01-14
Add hover popups to scores

fungjj92:
## Overview

To give a little more context to the apiary card content, scores will have popovers explaining the indicator and the score(s). The copy is yet to be provided.

Connects #372

### Demo

screen shot 2019-01-14 at 5 33 37 pm

screen shot 2019-01-14 at 5 33 55 pm

screen shot 2019-01-14 at 5 33 43 pm

### Notes

I could have broken the popover content into its own functional component but the apiary cards are already really nested and I didn't think doing so would add clarity.

## Testing Instructions

Run the project `./scripts/beekeepers.sh start`
Hover over an apiary card score


rajadain:
Wait, seeing this:

image

PR #421, opened 2019-01-15
Monthly Survey Form

rajadain:
## Overview

Adds a Monthly Survey form for taking in monthly surveys. Each form has a main `num_colonies` textbox, and then three tabs with the detailed forms.

The back-end endpoints were updated to allow sending multiple entries. Existing survey forms were updated accordingly.

Only those colonies with a "name" are saved. We should add a check to ensure there is at least one colony in the survey before saving it.

**TODO**

- [x] Add a check so that they have to fill out at least one colony
- [x] Add text to show that if the colony name is not entered, it will not be counted
- [x] Rebase on top of #416 to incorporate success modal
- [x] Make submission code match with April and November
- [x] monthly surveys should appear on the 15th of that month (i.e. July's survey appears on July 15)

Connects #381 

### Demo

image

### Notes

Lacks styling and validation, as do the other forms.

## Testing Instructions

* Check out this branch
* Run migrations `./scripts/manage.sh migrate`
* Start the app `./scripts/beekeepers.sh start`
* Log in to a "Pro" user account
* Go to surveys. Enter in Monthly surveys. Ensure they are saved correctly.
* Enter in November and April surveys. Ensure they still work.
* Load existing Monthly surveys. Ensure they are populated correctly.


src/icp/apps/beekeepers/views.py
```diff
@@ -60,23 +60,38 @@ def fetch_data(request):
...
+            "month_year": "042018",
+            "num_colonies": 3,
+            "survey_type": "APRIL", // or "NOVEMBER" or "MONTHLY"
+            "april": {              // Must be specified for "APRIL" surveys
+                "colony_loss_reason": ""
+            },
+            "november" : {          // Must be specified for "NOVEMBER" surveys
+                ...
+            },
+            "monthlies": [          // Must be specified for "MONTHLY" surveys
```
fungjj92: What was the reason to call this field "MONTHLIES" over "MONTHLY"?

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
```diff
@@ -0,0 +1,192 @@
...
+                colony_alive,
+                varroa_count_performed,
+            },
+        } = this.props;
+
+        const inputText = this.inputFactory('text');
+        const inputDate = this.inputFactory('date');
+        const inputNumber = this.inputFactory('number');
+
+        const colonyLossReasons = colony_alive ? null : (
```
fungjj92: nice easy solution to toggling fields conditionally!

src/icp/apps/beekeepers/views.py
rajadain: Because it's an array of (up to three) MONTHLY objects.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
```diff
@@ -0,0 +1,192 @@
...
+    VARROA_CHECK_METHODS,
+    MITE_MANAGEMENT_OPTIONS,
+} from '../constants';
+
+import { MonthlySurveyColony } from '../propTypes';
+
+class MonthlySurveyColonyForm extends Component {
+    inputFactory(inputType) {
+        const { data, onChange } = this.props;
+        const disabled = data.id !== null;
```
fungjj92: Not sure where to put this comment, so I'm putting it here randomly: 

I submitted a monthly survey successfully (200 from the endpoint) but after the fact, one of the fields highlighted orange. Have you seen this ? I don't know what that is supposed to indicate ? Might be a fluke, we can fix it in the future if it comes up.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+        const inputText = this.inputFactory('text');
+        const inputDate = this.inputFactory('date');
+        const inputNumber = this.inputFactory('number');
+
+        const colonyLossReasons = colony_alive ? null : (
+            <>
+                <div className="form__group">
+                    <label htmlFor="colony_loss_reason">
+                        What do you think the most likely cause of colony loss was?
```
fungjj92: I got an error submitting a monthly survey:
screen shot 2019-01-15 at 1 13 36 pm

Is this correct, that this field is required ? Not sure where the request is going wrong.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
fungjj92: hidden fields should have their values cleared out / set to default. this is an issue if the hidden fields were filled out then toggled off.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
```diff
@@ -51,18 +155,120 @@ class MonthlySurveyForm extends Component {
...
+            num_colonies,
+            apiary,
+            month_year,
+            survey_type: SURVEY_TYPE_MONTHLY,
+            monthlies,
+        };
+
+        getOrCreateSurveyRequest({ apiary, form })
+            .then(() => {
+                // TODO Replace with success message
```
fungjj92: See the PR for #419 🤔We'll have to see what we do here -- parse the returned data back to the form, or close the modal.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
```diff
@@ -51,18 +155,120 @@ class MonthlySurveyForm extends Component {
...
+            // Skip if colony name was not specified
+            if (!monthly.colony_name) {
+                return null;
+            }
+
+            const multipleChoiceData = {};
+
+            this.multipleChoiceKeys.forEach((mcKey) => {
+                const keyOptions = Object.entries(monthly)
+                    .filter(([k, v]) => v && k.startsWith(mcKey))
```
fungjj92: Within this same function, there's some differently named variables than in April and Nov survey components, e.g. `mcKey`. Purposeful ? Worthwhile to get them identical for future readings of this, if indeed they are, and perhaps moved in utils? Ideally we'd have a parent component they all inherit from, but we've talked about pushing that off 😢

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
```diff
@@ -0,0 +1,373 @@
...
+                            className="form__control"
+                            id="num_colonies"
+                            name="num_colonies"
+                            onChange={this.handleChange}
+                            value={num_colonies}
+                            disabled={completed}
+                            required
+                        />
+                    </div>
+                    Please enter details for up to three colonies:
```
fungjj92: Upon viewing a completed Monthly survey, would be nice to hide this entire section or replace with alternate text "No colony data entered" or something if there's no colony data.

Currently looks like:
screen shot 2019-01-15 at 1 53 23 pm

rajadain:
`num_colonies` controlled / uncontrolled issue fixed in 99ee7b59

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
rajadain: Fixed in bfdf06c4

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
rajadain: I'm going to make it so that you have to enter at least one colony.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
rajadain: I think this syntax is a little cleaner, so will update the other two to use this so it's all consistent.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
fungjj92: Fixed in #419 ~ verdict: close the modal and show the success modal !

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
rajadain: Done in 0feff3a

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
rajadain: Done in b2bf257

rajadain:
Okay, this is ready for another look!

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
```diff
@@ -0,0 +1,430 @@
...
+        super(props);
+        this.initialMonthly = {
+            id: null,
+            inspection_date: '',
+            colony_name: '',
+            colony_alive: false,
+            colony_loss_reason: '',
+            num_bodies_supers_deep: 0,
+            num_bodies_supers_medium: 0,
+            num_bodies_supers_shallow: 0,
```
fungjj92: The fields whose starting value are 0 and inputs have type `number` have a kinda strange default behavior. With a starting value 0, you can't backspace / clear the field --  only scroll from 0 or like..prepend 0 with another number i.e put a 9 in from of 0 --> 90. That's weird. The other similar inputs, like `num_colonies` have starting value `''` -- which is obvi not of type `number` but makes the input interaction logical.

ie.
screen shot 2019-01-18 at 9 44 55 am

src/icp/apps/beekeepers/js/src/utils.js
```diff
@@ -258,16 +258,23 @@ export function getMonthlySurveys({ id: apiary, created_at: createdAt, surveys }
...
-            const survey = matchSurvey(month_year) || {
-                apiary,
-                month_year,
-                survey_type,
-                completed: false,
-            };
-
-            mthSurveys.push(survey);
+            // Don't list current month until the 15th
+            if (!(
```
fungjj92: i've learned that your brain prefers not statements more than mine does. this condition is hard for me to read. is a positive statment where the `date > 14` as clear to you?

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
rajadain: If I give it an initial value of `''`, I get the following error:

image

This can be alleviated by switching it to a `text` type rather than `number`, but then we lose the built-in validation that disallows non-numeric input. I think we can let it be for now.

src/icp/apps/beekeepers/js/src/utils.js
rajadain: The negation is not for the date comparison, but for the group. The logical equivalent is this:

```js
if (
    y !== endYear
    || m !== endMonth
    || currentDate.getDate() >= 15
)
```

is this clearer to you?

rajadain:
Thanks for reviewing! Merging this now.

Issue #414, opened 2019-01-10
Persist map center between page views


mmcfarland:
When apiaries exist on the Location Finder tab, and one cycles between the Survey and Location Finder tab, the previous map center/zoom is lost.  

The settings should be persisted or at least elastic to the bounds of the apiaries, as they will likely now be offscreen.

Issue #391, opened 2018-12-20
Sort options in dropdown don't match apiary card


fungjj92:
Since the forage raster was renamed to floral, there's a mismatch in the front end. I think we should stick to one naming convention and be firm about it.

<img width="464" alt="screen shot 2018-12-20 at 5 05 10 pm" src="https://user-images.githubusercontent.com/10568752/50313461-727f2580-0479-11e9-8f9e-91c19000806b.png">

dboyer:
We should go with Floral, per feedback from Penn State.

PR #428, opened 2019-01-21
Mixed bag of bugs

fungjj92:
## Overview

Small clean ups that didn't warrant individual PRs
- Refreshing or reopening the page should load the map where the user last was, rather than initializing the map to the default location in PA. This makes more sense from the user's perspective. For this, I used redux persist! It finally came into handy.
- Forage rasters were renamed Floral rasters, but the new wording wasn't *official* until now. Make the resulting copy changes.

Connects #414, #391

### Demo

map persist

screen shot 2019-01-21 at 4 36 29 pm

### Notes

I considered initializing the map at the bounds of the apiaries but I prefer this UX and the alternative felt overly complicated.

## Testing Instructions

Move the map and refresh or open a new page -- make sure the map opens at the last spot.

See the sort dropdown works and the text is correct

### DIFF
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -40,6 +40,7 @@ export const openEmailFormModal = createAction('Open email form modal');
 export const closeEmailFormModal = createAction('Close email form modal');
 export const openSuccessModal = createAction('Open success message modal');
 export const closeSuccessModal = createAction('Close success message modal');
+export const setMapCenter = createAction('Set map center');
 
 
 export function fetchApiaryScores(apiaryList, forageRange) {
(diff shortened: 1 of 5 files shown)


rajadain:
It should still say "Forage Range", and not "Floral Range", right?

fungjj92:
The good news is that postgres must add quotes when appropriate. Here is the output of april surveys when a string entry contains commas (last row).
```
survey_id,colony_loss_reason
55,varroa mites
60,varroa mites
63,INADEQUETE_FOOD_STORES;OTHER-super sad bee killer
64,VARROA_MITES;INADEQUETE_FOOD_STORES;OTHER-super sad bee killer
65,VARROA_MITES;OTHER-super sad bee killer2
66,COLONY_TOO_SMALL;PESTICIDE_EXPOSURE;OTHER-
67,VARROA_MITES;POOR_WEATHER_CONDITIONS;OTHER-
69,"VARROA_MITES;OTHER-super, sad, bee, killer"
```

PR #457, opened 2019-01-30
Support a second domain for staging and production

### LINKED ISSUE #427: Domain name
The main domain is beescape.org with beescape.net and beescape.info also purchased

Requires #432

jeancochrane:
## Overview

In addition to the existing `{staging}.app.pollinatormapper.org` domains, support the following two domains:

- `app.beescape.org`
- `staging.app.beescape.org`

Update CloudFormation resources to reflect this change. In addition, update a few references to the legacy landing page in the application code.

Helps resolve #427.

### Notes

* This PR is blocked until the DNS manager has added an NS record to point to our hosted zone.

## Testing Instructions

 * _Testing TK once PR is unblocked_


README.md
```diff
@@ -2,9 +2,9 @@
 
 ## URLs
 
-Staging: https://staging.app.pollinationmapper.org
```
hectcastro: For these documentation references, I think we want both domains to be represented vs. swapping one for the other.

deployment/RELEASES.md
```diff
@@ -34,7 +34,7 @@ If you cannot see columns for `Branch`, `Environment`, and `Service`, use the ge
 
 ## Release Testing
 
-After the `staging-deployment` job completes, `staging.app.pollinationmapper.org` should reflect the current release. Be sure to run any outstanding database migrations or data imports.
```
hectcastro: Same comment here.

src/icp/js/src/draw/templates/splash.html
```diff
@@ -11,7 +11,7 @@ <h2>Compare pollination management scenarios</h2>
     <li>Compare how different scenarios impact your crop’s pollination and yield.</li>
 </ol>
 <p>
-This <a target="_blank" rel="noopener noreferrer" href="http://pollinationmapper.org"><b><i class="fa fa-book"></i> Help Page</b></a> will walk you through the process in more detail.
```
hectcastro: Might be a better question for @rajadain here. Do we leave these as is or add functionality to make them dynamic?

src/icp/js/src/modeling/tours.js
```diff
@@ -42,7 +42,7 @@ var newScenarioTour = createTour(
         steps: [
             {
                 target: "add-pollinator-planting-btn",
-                content: 'Select a pollinator planting type from this menu, then draw it on the scenario map. Plantings can be drawn outside of the selected farm or field area, but cannot be stacked on top of each other. The minimum planting size is 100 ft &times; 100 ft. See <a target="_blank" rel="noopener noreferrer" href="http://pollinationmapper.org")>help guide</a> for more information.',
```
hectcastro: Similar comment here.

cc: @rajadain

src/icp/js/src/draw/templates/splash.html
rajadain: In this case we should leave the existing domain, as this page is only every visible from the Pollination Mapper app. All the Beekeepers front- and back-end code is in [`src/icp/apps/beekeepers`](https://github.com/project-icp/bee-pollinator-app/tree/develop/src/icp/apps/beekeepers), so the content outside of it should not be affected.

src/icp/js/src/modeling/tours.js
rajadain: Same comment as above, the URL here should not be changed.

Issue #468, opened 2019-02-08
Add Score Information in Popups


rajadain:
Use the following table to populate the popup to indicate to the user which bucket their score falls under:

<img width="366" alt="image" src="https://user-images.githubusercontent.com/1430060/52492579-79800580-2b97-11e9-8ca5-b61b6de049a9.png">

![image](https://user-images.githubusercontent.com/1430060/52492688-afbd8500-2b97-11e9-9617-2bc3a4f0602a.png)

Show the entire row for the appropriate type. E.g. if the type is Spring Floral, we show a table of the five stops and scores for Spring Floral.

Optionally, highlight / make prominent their score in the table or popup description.

rajadain:
Client Notes:

> How to read this:
> 
> Anything at the low score or lower is considered “low levels” (40 or less for spring)
>
> Anything at the high score or higher is considered “high levels” (53 or more for spring)
> 
> I think Azavea converted the pesticide index to a 1-100 scale, so the scores should be converted accordingly.
> 
> High is good for floral resources and nesting, BAD for pesticide load.  This may be confusing for people (at least, an assistant prof working with us was confused by this…). Might be good to use some color coding?

rajadain:
@mmcfarland Is it true that we changed the Pesticide scale to 1-100? Or are we just multiplying the values when rendering in the front-end? As far as I can tell we're only multiplying the displayed results with 100, same as for every other indicator:

![image](https://user-images.githubusercontent.com/1430060/52506057-49e2f480-2bbb-11e9-926b-e6e0c12be0c2.png)

![image](https://user-images.githubusercontent.com/1430060/52506094-63843c00-2bbb-11e9-92d2-0c0911e2c796.png)

and haven't rescaled the original values.

dboyer:
@rajadain  Text is:

Spring Floral Quality:
High >53
Low <40

Summer Floral Quality
High >49
Low <44

Fall Floral Resource Quality
High >36
Low <30

Nesting Quality  
High >46 
Low <33

Pesticide 
High Expected Levels >37 
Low Expected Levels <13

mmcfarland:
> and haven't rescaled the original values.

@rajadain Sorry I missed this. That is correct.

dboyer:
@rajadain Penn State confirmed that the remaining sentence of lorem ipsum can just be removed from each of the pop-up boxes. There does not need to be any type of summary sentence in the box.

PR #471, opened 2019-02-08
Scores in Label Tooltips

rajadain:
## Overview

Adds score ranges, classification in the score popup.

This could potentially use some design input.

Connects #468 

### Demo

image

Tagging @dboyer for visual review.

### Notes

Need to confirm the Pesticide scores, which we are showing differently than their scale.

## Testing Instructions

* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and add some apiaries
* Hover over their scores.
  - Ensure you see the table of expected results in the popup
  - Ensure the scores and their classification is in line with the description in #468

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
@@ -38,20 +38,22 @@ const ScoresLabel = ({ indicator, scores }) => {
                 <table className="scoreTable">
                     <thead>
                         <tr>
-                            <th className="scoreTable__th">Min</th>
                             <th className="scoreTable__th">Low</th>
-                            <th className="scoreTable__th">Med</th>
+                            <th className="scoreTable__th">Medium</th>
                             <th className="scoreTable__th">High</th>
-                            <th className="scoreTable__th">Max</th>
                         </tr>
                     </thead>
                     <tbody>
                         <tr>
-                            <td className="scoreTable__td">{indicatorDetails.scoreStops[0]}</td>
-                            <td className="scoreTable__td">{indicatorDetails.scoreStops[1]}</td>
-                            <td className="scoreTable__td">{indicatorDetails.scoreStops[2]}</td>
-                            <td className="scoreTable__td">{indicatorDetails.scoreStops[3]}</td>
-                            <td className="scoreTable__td">{indicatorDetails.scoreStops[4]}</td>
+                            <td className="scoreTable__td">
+                                {`≤ ${indicatorDetails.scoreStops[1]}`}
+                            </td>
+                            <td className="scoreTable__td">
+                                {indicatorDetails.scoreStops[2]}
+                            </td>
+                            <td className="scoreTable__td">
+                                {`${indicatorDetails.scoreStops[3]} ≤`}
+                            </td>
                         </tr>
                     </tbody>
                 </table>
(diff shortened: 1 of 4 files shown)


src/icp/apps/beekeepers/js/src/utils.js
```diff
@@ -313,3 +313,23 @@ export function getOrCreateSurveyRequest({ apiary, id, form }) {
...
+ *
+ * @param {number} score
+ * @param {array} stops
+ */
+export function classifyScore(score, stops) {
+    if (score < stops[1]) {
+        return 'Low';
+    }
+
+    if (score < stops[3]) {
```
fungjj92: I imagine these should be `<=`

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
```diff
@@ -35,6 +35,31 @@ const ScoresLabel = ({ indicator, scores }) => {
...
+                            <td className="scoreTable__td">{indicatorDetails.scoreStops[1]}</td>
+                            <td className="scoreTable__td">{indicatorDetails.scoreStops[2]}</td>
+                            <td className="scoreTable__td">{indicatorDetails.scoreStops[3]}</td>
+                            <td className="scoreTable__td">{indicatorDetails.scoreStops[4]}</td>
+                        </tr>
+                    </tbody>
+                </table>
+                <div className="scoreVerdict">
+                    {score}
+                    {' = '}
```
fungjj92: I'll defer to design for styling this. My inclination is the scores should be bold/heavier font-weight but not necessary to change on an inkling.

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
```diff
@@ -35,6 +35,31 @@ const ScoresLabel = ({ indicator, scores }) => {
         >
             <div className="indicator__popup">
                 <div className="name">{indicatorDetails.name}</div>
+                <table className="scoreTable">
+                    <thead>
+                        <tr>
+                            <th className="scoreTable__th">Min</th>
```
fungjj92: What does `min` mean? Is it the lowest score in the raster ? Minimum score to be viable for bees to live? Same with `max`. Either we need that in the text descriptionsthe client will send, or the min/max should be 0 and 100 respectively.

rajadain: Hmm, good read, that makes sense. I'll drop the `Min` and the `Max`, and change the UI to indicate that `Low` and lower would be `Low`, and `High` and higher would `high`.

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
rajadain: > but not necessary to change on an inkling

Can you rephrase this? I don't follow.

fungjj92: Just that I'll defer styling opinions to design

src/icp/apps/beekeepers/js/src/components/ScoresLabel.jsx
rajadain: Is this any better?

image

fungjj92: Worth presenting to beekeepers in the next test phase ~

dboyer:
This looks good for now. I've sent the screenshot to Penn State to provide the text replacement for the lorum ipsum.

Issue #386, opened 2018-12-17
Create Data Upload IAM user


mmcfarland:
Admin users will be able to upload updated rasters to the site over time.

* [ ] Create an IAM user that is scoped for write access to the `beekeepers-data-us-east-1` only.
* [ ] Take screenshots and light documentation on the steps to upload raster files

PR #478, opened 2019-02-14
Data upload docs

mmcfarland:
## Overview

Add documentation for updating beekeeper rasters

Connects #386 

### Notes

The `beekeeper-data` credentials are in LastPass

## Testing Instructions

 * Confirm you can log in with the referenced IAM account, and that the account _only_ has access to the beekeepers raster s3 bucket, and no other services/buckets.
 * Confirm the steps listed in the docs are correct.


rajadain:
Taking a look now.

rajadain:
When trying to upload a test file following the docs, I just get this message in the AWS UI:

image

with no detail about why it failed. This is the file I'm trying to upload:

mmcfarland:
Thank you for catching that, the UI was not obvious that the error was happening, but I do get the same --- appears to be insufficient privileges for the user.

screenshot from 2019-02-18 11 57 07

rajadain:
I just tried this again. If I don't uncheck "Object Read" and "Object Write" permissions, it works, otherwise it still fails.

mmcfarland:
I updated the docs and screenshot for the step, to include Object Read/Write.

Issue #486, opened 2019-03-07
Text changes based on user feedback


dboyer:
- [x] On Registration survey, change "Do you buy queens, nucs, or packages..." to "Do you regularly buy queens, nucs...."

- [x] Add note on resend activation link and registration page saying "Please check your email for an account activation link. The email may go to your spam folder, depending on your email settings."

- [x] On November survey, change "How much honey did you collect on average for each colony" to "Approximately how much honey did you collect for each colony this year?"

- [x] ~Change "Pesticide quality" to "Insecticide load". This will also mean changing the Pesticide label to Insecticide~ Moved to #488 

- [x] In user registration survey, add "Warre hive" as an option under the "what kind of equipment do you use" question

- [x] In user registration survey, change the wording of the income from bees question to read "Do you receive income from any of the following? Check all that apply."

- [x] On monthly survey, change answers to "what methods of mite management did you use?" to this list: None, formic acid, oxalic acid, thymol (apilife or apivar), amitraz (apivar), apistan, checkmite,  queen manipulation, drone remove, other

- [x] On November survey, change answers to "what methods of mite management did you use?" to this list: None, formic acid, oxalic acid, thymol (apilife or apivar), amitraz (apivar), apistan, checkmite,  queen manipulation, drone remove, other

- [x] On monthly survey, add "Bear/natural disaster" as to choice for "What do you think was the most likely cause of colony loss?"

- [x] Home should link to beescape.org and Methodology should link to beescape.org/methodology

- [x] Add link to wild bee hotel survey - http://beescape.org/wild-bees/  - in the text that displays on the survey page when the user is not logged in 

- [x] On the user survey, change the "Describe their genetics" question to "What is the stock or genetic background of your queens?"

- [x] On the user survey, change the answers for "How do you decide when to manage for Varroa?" from a text field to a multiple-choice including "mite counts, mite symptoms, time of year, or other - fill in the blank"

fungjj92:
I spun off one of the bullets as #488 because it will require a more follow through and commitment from the client.

Also, for mite management naming I am creating tooltips to capture the references to produces i.e apivar, apilife instead.

fungjj92:
@dboyer  Please run these tooltip descriptions by the client. I did research on these chemicals and products and this is the conclusion I came to.

[...]

PR #489, opened 2019-03-07
Text changes for forms from beekeepers feedback

fungjj92:
## Overview

The app was tested by an elite group of beekeepers and these are the lower hanging fruit text changes.

Connects #486

### Demo

screen shot 2019-03-07 at 2 15 57 pm
screen shot 2019-03-07 at 2 16 00 pm
screen shot 2019-03-07 at 2 16 30 pm
screen shot 2019-03-07 at 11 31 58 am
screen shot 2019-03-07 at 11 32 20 am
screen shot 2019-03-07 at 3 17 49 pm

### Notes

I made a judgment call on the thymol and amitraz keys to turn the extra context into a tooltip, but I also want the descriptions I then wrote run by the client.

In the interest of saving time I didn't do the refactoring in forms that I would have otherwise liked 😢 

## Testing Instructions

Make sure that all the surveys open and can submit successfully.
No typos.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
@@ -44,6 +44,7 @@ class UserSurveyModal extends Component {
             equipment_8_frame: '',
             equipment_10_frame: '',
             equipment_top_bar: '',
+            equipment_warre_hive: '',
             equipment_other: '',
         };
         // If a `field` has the `value`, then all the fields in `reset` are set
@@ -128,6 +129,7 @@ class UserSurveyModal extends Component {
             equipment_8_frame,
             equipment_10_frame,
             equipment_top_bar,
+            equipment_warre_hive,
             equipment_other,
         } = this.state;
 
@@ -136,7 +138,13 @@ class UserSurveyModal extends Component {
         );
         const prepended_equipment_other = equipment_other.length ? `OTHER-${equipment_other}` : '';
         const equipment = arrayToSemicolonDelimitedString(
-            [equipment_8_frame, equipment_10_frame, equipment_top_bar, prepended_equipment_other],
+            [
+                equipment_8_frame,
+                equipment_10_frame,
+                equipment_top_bar,
+                equipment_warre_hive,
+                prepended_equipment_other,
+            ],
         );
         const form = Object.assign({}, this.state, {
             income,
@@ -177,6 +185,7 @@ class UserSurveyModal extends Component {
             equipment_8_frame,
             equipment_10_frame,
             equipment_top_bar,
+            equipment_warre_hive,
             equipment_other,
         } = this.state;
 
@@ -315,8 +324,7 @@ class UserSurveyModal extends Component {
                             </div>
                             <div className="form__group">
                                 <label htmlFor="income">
-                                    Do you obtain income from your bees?
-                                    What do you receive income from?
+                                    Do you receive income from any of the following?
                                     Check all that apply.
                                 </label>
                                 <div className="form__checkbox">
@@ -429,7 +437,7 @@ class UserSurveyModal extends Component {
                             )}
                             <div className="form__group">
                                 <label htmlFor="purchased_queens">
-                                    Do you buy queens, nucs or packages?
+                                    Do you regularly buy queens, nucs or packages?
                                 </label>
                                 <select
                                     className="form__control"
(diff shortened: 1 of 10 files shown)


src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
```diff
@@ -315,8 +324,7 @@ class UserSurveyModal extends Component {
                             </div>
                             <div className="form__group">
                                 <label htmlFor="income">
-                                    Do you obtain income from your bees?
-                                    What do you receive income from?
+                                    Do you receive income from any of the following?
```
rajadain: Should we remove `NO_INCOME` from this, or use it as a default, given this new phrasing?

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -160,3 +159,13 @@ export const QUEENRIGHT_DESCRIPTION = {
     title: '',
     body: 'You can determine if a queen is present by locating her physically or checking for eggs. However, by late fall the queen stops laying eggs. We ask that beekeepers use their best judgement when making the assessments during these times.',
 };
+
+export const THYMOL_DESCRIPTION = {
+    title: '',
+    body: 'The active ingredient in Apilife Var is thymol.',
```
rajadain: Is "Apilife Var" a mistake? The issue says "apilife or apivar" as if they're brand names.

src/icp/apps/beekeepers/js/src/constants.js
```diff
@@ -59,17 +59,15 @@ export const SURVEY_TYPE_MONTHLY = 'MONTHLY';
...
-    'CHEMICAL_OXALIC_ACID_VAPORIZATION',
-    'CHEMICAL_OXALIC_ACID_DRIBBLE',
-    'CHEMICAL_THYMOL_MENTHOL_APILIFE',
-    'CHEMICAL_THYMOL_MENTHOL_APIGUARD',
-    'CHEMICAL_SYNTHETIC_APIVAR',
-    'CHEMICAL_SYNTHETIC_APISTAN',
-    'CHEMICAL_SYNTHETIC_CHECKMITE_PLUS',
-    'MECHANICAL_DRONE_BROOD_REMOVAL',
-    'MECHANICAL_QUEEN_MANIPULATION',
+    'THYMOL',
```
rajadain: The loss of these prefixes means that they won't easily be able to group "Chemical", "Mechanical", and other classes. It would be better to keep an internal representation, and have a dictionary that mapped it to labels, so we could change the labels without changing the identifiers. This also allows renaming labels in the future, without changing the underlying identifier.

But since the app hasn't launched  yet, I think this change is harmless at this point. Furthermore, there aren't that many categories, so they can still group them manually after the fact.

Thus, we don't need to change this. This is just an observation.

rajadain: On another note, we should update these lists to match:

https://github.com/project-icp/bee-pollinator-app/blob/3a3f12c11ca17847be9679e8914120ddd3373848/src/icp/apps/beekeepers/models.py#L213-L220

https://github.com/project-icp/bee-pollinator-app/blob/3a3f12c11ca17847be9679e8914120ddd3373848/src/icp/apps/beekeepers/models.py#L328-L335

src/icp/apps/beekeepers/js/src/constants.js
fungjj92: I had posted this to the issue to run by the client.

I had done research on all these unfamiliar words and the captions I came up with is what I learned. It struck me as odd that Apivar was listed under both chemicals and turned out theres a product also called Apilife Var too (so confusing!) so I thought maybe the client made an error.

rajadain: Ah I see. Okay, let's hold off on merging this until we hear back then.

src/icp/apps/beekeepers/js/src/components/UserSurveyModal.jsx
fungjj92: Hm. Yea, that makes sense.

fungjj92:
I've pushed up changes. Feel free to merge this in my absence.

src/icp/apps/beekeepers/js/src/constants.js
rajadain: This has been resolved in ec7e1fa.

rajadain:
@mmcfarland I feel pretty confident about merging these text changes in, but if you could take a quick look and 👍 that'd be great.

Issue #487, opened 2019-03-07
Add two new data fields


dboyer:
- [x] On November survey, add the question "Did you treat all of the colonies in the apiary?" with answers of Yes/No

- [x] On the monthly survey, add the question "If you have an automated scale associated with this colony, please enter the hive scale ID number here." and then display a text box

rajadain:
Do they have a preference for where to show these questions in the survey? Beginning / middle / end?

rajadain:
Also, for the first question "Did you treat all of the colonies in the apiary?", what should be the default value applied to current accounts? I don't think we can reset the staging database as Beekeepers have been testing there, so we need to initialize this new field to either `True` or `False` for the November Surveys that have already been filled out.

dboyer:
Let's put automated scale under the number of hive bodies and supers, right before Varroa count.

All the staging data will not be ported to production, right? In that case, setting them all to "true" is fine.

dboyer:
@rajadain See above edit about the placement of a question

PR #491, opened 2019-03-08
Add New Questions to Surveys

rajadain:
## Overview

Adds a new field each to November and Monthly surveys. This includes a model change, database migration, and additions to the respective React components.

Connects #487 

### Demo

image

image

image

### Notes

I added the new fields to the very bottom of the current set of questions. For the new November Survey question "Did you treat all of the colonies in the apiary?", the default is "Yes". If these assumptions need to change, please let me know. /cc @dboyer 

## Testing Instructions

* On `develop`, go to [:8000/?beekeepers](http://localhost:8000/?beekeepers), log in, and create some apiaries. Fill out a Monthly and a November survey.
* Check out this branch and `manage migrate`. Ensure migrations run successfully.
* Go back to your surveys. Ensure existing surveys have reasonable default values for the new fields.
* Fill out new surveys and ensure they are saved correctly.

### DIFF
--- a/src/icp/apps/beekeepers/models.py
+++ b/src/icp/apps/beekeepers/models.py
@@ -222,6 +222,10 @@ class NovemberSurvey(models.Model):
         null=True,
         help_text='What methods of mite management do you use? '
                   'Select all that apply.')
+    all_colonies_treated = models.BooleanField(
+        null=False,
+        default=True,
+        help_text='Did you treat all of the colonies in the apiary?')
 
     def __unicode__(self):
         return unicode('{}:{}:{}'.format(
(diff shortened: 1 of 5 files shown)


src/icp/apps/beekeepers/models.py
```diff
@@ -221,6 +221,10 @@ class NovemberSurvey(models.Model):
         null=True,
         help_text='What methods of mite management do you use? '
                   'Select all that apply.')
+    all_colonies_treated = models.BooleanField(
+        null=False,
+        default=True,
```
fungjj92: Was this the requested default? Otherwise I'd think to default to `False`.

rajadain: There was no requested default, I'm still awaiting confirmation (see https://github.com/project-icp/bee-pollinator-app/issues/487#issuecomment-470599030).

I went with `True` because before this field existed, I imagine the assumption was the data applied to all the colonies. I'll hold off on merging this until we have an explicit decision.

rajadain: Got confirmation that `True` should be the default here: https://github.com/project-icp/bee-pollinator-app/issues/487#issuecomment-471697476

rajadain:
Rebased on `develop` to resolve conflicts. Will merge when green. Thank you for taking a look!

Issue #488, opened 2019-03-07
Update 'pesticide' to 'insecticide load'


fungjj92:
Change "Pesticide quality" to "Insecticide load". This will also mean changing the Pesticide naming convention to Insecticide, renaming the underlying COGs and updating the `vrt` accordingly. Confirm with the client that this is what they want and can update data correctly in the future.

Note that renaming files on s3 will break staging. Create duplicates of pesticide data, renamed to insecticide, and new corresponding `vrts`. Be sure to delete stale data off s3 once this work is merged and deployed to staging.

Ensure production data, if we have production at the time, is updated as necessary.

dboyer:
Confirmed that we can change the naming convention to Insecticide load. Penn State will adjust the name of any insecticide rasters they upload in the future.

mmcfarland:
I'm curious if this can be done by simply updating the label for Pesticide without having to update all of our raster sources and scripts? Perhaps with a comment if the concern is that it will be confusing in the future for which dataset represents Insecticide?

PR #494, opened 2019-03-12
Rename Pesticide quality to Insecticide load

rajadain:
## Overview

This affects only the labels and not the inner variable definitions or S3 raster names. Thus, the upload instructions for the clients do not need to be updated.

Connects #488 

### Demo

image

## Testing Instructions

* Go to [:8000/?beekeepers](http://localhost:8000/?beekeepers) and add an apiary
* Ensure the scores say "Insecticide" and "Insecticide load" instead of "Pesticide" or "Pesticide quality"


rajadain:
As part of this implementation, we do _not_ need to change the upload instructions. /cc @dboyer

rajadain:
Ah good catch. That should now be fixed in 36f47a5:

image

In the interest of expediency, I switched to using the long names used inside the score labels. This makes the dropdown just a bit wider than before, but I tested on my 13" laptop and it didn't cause any spillover.

Issue #490, opened 2019-03-07
Custom font doesn't load in IE11


fungjj92:
IE is not recognizing the custom font, and instead defaulting to one of the fall backs (arial, I think).

Edit: I looked into this, my findings and passing the baton:

Nothing jumped out at me as an issue.

Our webpack config is pretty up to date, and includes the css `autoprefixer` plugin that usually solves IE compatibility issues with css. Our custom font works in all browsers including Edge, just not IE11. I was unable to figure out the source of our font (i.e. is it a Google font?). 

We include all font file versions (`.eot` is the format that IE11 requires) and import them correctly (`@import(...)`).

In my research I came across a new setting in Windows 10  that blocks suspicious font downloading, but I've dispelled this idea by confirming fonts will don't work on IE11 on Windows 8.1 on Browserstack. https://docs.microsoft.com/en-us/windows/security/threat-protection/block-untrusted-fonts-in-enterprise#Turn_on_and_use_the_Blocking_untrusted_fonts_feature

I did not get time to evaluate this potential clue: https://github.com/facebook/create-react-app/issues/4121

Issue #465, opened 2019-02-04
Survey Page Scroll Bug in Firefox


rajadain:
The light background stretches all the way to the screen, but no further, and when scrolling down reveals a darker gray:

![2019-02-04 17 23 05](https://user-images.githubusercontent.com/1430060/52241460-bdc28b80-28a1-11e9-990d-66082572d85f.gif)

rajadain:
Shares points with #495

PR #495, opened 2019-03-13
Various IE11 Fixes

rajadain:
## Overview

Fixes the font loading issue in IE11, using ideas from [here](https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f). We should consider flipping this sequence for all projects to ensure IE11 compatibility.

Also fixes tall modals in IE11 which were previously transparent below the window:

image

Also fixes the November Survey to load correctly in read-only mode. Previously some fields were not initialized correctly, and dropdowns were not disabled.

Connects #490 
Connects #465 

### Demo

Fonts rendering correctly:

image

Modals scrolling correctly:

image

November Survey reinitializing correctly:

image

## Testing Instructions

* Check out this branch and `beekeepers build`
* Test the site in IE11 (BrowserStack or Windows VM)
  - [x] Ensure the fonts load correctly
  - [x] Ensure that tall modals (like surveys) scroll and look correct
* Fill out a November Survey and save it. Make sure you pick non-default choices for all dropdowns.
* Load the same survey to view it in read-only mode.
  - [x] Ensure all the fields are correctly filled
  - [x] Ensure all fields are disabled

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/NovemberSurveyForm.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/NovemberSurveyForm.jsx
@@ -384,6 +384,7 @@ class NovemberSurveyForm extends Component {
                         value={varroa_manage_frequency}
                         onChange={this.handleChange}
                         className="form__control"
+                        disabled={!!completedSurvey}
                     >
                         <option value="NEVER">I did not</option>
                         <option value="ONCE">Once a year</option>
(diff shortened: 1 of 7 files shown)


rajadain:
Good catch. Fixed in 8480277

image

Also found and fixed a IE 11 that was previously only seen in Firefox #465 in eff9d3b.

This is ready for a final look.

rajadain:
@azavea-bot rebuild

PR #500, opened 2019-03-18
Fix Staging Issues

rajadain:
## Overview

Fixes issues found during staging testing.

### Demo

Previously this button was overlapping content:

image

That is now fixed:

image

---

Previously reloading the main app after logging in would fetch apiaries in a random order:

image

They are now always fetched in the correct order:

image

## Testing Instructions

* Check out this branch and log in
* Make sure you have 0 apiaries. Delete all of them if needed.
* Go to the Survey page.
  - [x] Ensure the button pointing to the Location Finder isn't overlapping content
* Go to the Location finder.
* Add some apiaries. Delete one or two from the middle (not the first or the last).
* Refresh the page.
  - [x] Ensure the apiaries show up in the correct ordering for "Default"
  - [x] Ensure you can delete existing apiaries, and the correct ones are deleted
  - [x] Ensure you can add new apiaries, and they have the correct marker assigned

### DIFF
--- a/src/icp/apps/beekeepers/sass/06_components/_survey.scss
+++ b/src/icp/apps/beekeepers/sass/06_components/_survey.scss
@@ -44,6 +44,7 @@
     }
 
     &__button {
+        display: inline-block;
         max-width: max-content;
         margin: 2rem 0;
         padding: 15px;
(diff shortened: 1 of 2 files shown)


rajadain:
CI failed due to leftover `libpq-dev` issues, not a real problem. Going to merge this now.

Issue #463, opened 2019-02-04
Add monthly survey field and autofill functionality


dboyer:
- "Do you have an automated scale associated with this colony?" added as a required question for each colony in the monthly survey
- If the user says yes, they would then be prompted to enter the brand and ID number as a required field
- If they say yes, the entered brand and ID number are autofilled on the following monthly surveys
- Colony name also becomes a field autofilled on the following monthly surveys
- Beekeepers can edit the colony name and brand and ID number for the automated scale if needed

rajadain:
Automated Scale ID field was added in #491.

The colony name and automated scale ID field currently do not autofill, and need to be entered each time.

dboyer:
We do not need to autofill at this time. I've labeled this as an enhancement.

rajadain:
Blocked until confirmed by client

PR #516, opened 2019-06-10
Autofill select fields in monthly surveys

fungjj92:
## Overview

Adds autofill using info from the most recent completed monthly survey.

Connects #463 

### Demo

Screen Shot 2019-06-10 at 5 13 51 PM

Screen Shot 2019-06-12 at 1 52 47 PM

Completed surveys:
Screen Shot 2019-06-12 at 5 43 00 PM

Autofilled surveys in id order:
Screen Shot 2019-06-12 at 5 42 54 PM

## Notes
There was an existing console warning about un/controlled inputs. There's a new console warning about requiring the `month_year` field on `lastMonthlySurvey` in some components that I haven't figured out how to clean up 🤔 

Screen Shot 2019-06-10 at 5 23 44 PM

I rolled in #517

## Testing Instructions
- [ ] Run migrations
- [x] Successfully submit a monthly survey on an apiary without monthly surveys
Start w a blank form
- [x] Submit a subsequent monthly survey and see the hive and colony name info autofill, but change the autofilled info
- [x] See that subsequent monthly surveys use the more recent month data
- [ ] Check the network tab that the opening any type of survey only sends the expected requests

- [ ] Check that monthly form submissions work in Safari and IE11

Connects #463, #517

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/ApiarySurveyListing.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/ApiarySurveyListing.jsx
@@ -4,13 +4,14 @@ import { arrayOf } from 'prop-types';
 import { Apiary, Survey } from '../propTypes';
 import SurveyCardListing from './SurveyCardListing';
 
-const ApiarySurveyListing = ({ apiary, surveys }) => {
+const ApiarySurveyListing = ({ apiary, surveys, lastMonthlySurvey }) => {
     const { name } = apiary;
     const listings = surveys.map(survey => (
         <SurveyCardListing
             key={name + survey.month_year + survey.survey_type}
             apiary={apiary}
             survey={survey}
+            lastMonthlySurvey={lastMonthlySurvey}
         />
     ));
 
@@ -22,9 +23,14 @@ const ApiarySurveyListing = ({ apiary, surveys }) => {
     );
 };
 
+ApiarySurveyListing.defaultProps = {
+    lastMonthlySurvey: null,
+};
+
 ApiarySurveyListing.propTypes = {
     apiary: Apiary.isRequired,
     surveys: arrayOf(Survey).isRequired,
+    lastMonthlySurvey: Survey,
 };
 
 export default ApiarySurveyListing;
--- a/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
@@ -177,7 +177,7 @@ class MonthlySurveyColonyForm extends Component {
             <>
                 {inputText('colony_name', 'Colony Name (required)', required)}
                 {inputDate('inspection_date', 'Date of Inspection', required)}
-                {inputText('hive_scale_id', 'If you have an automated scale associated with this colony, please enter the hive scale ID number here.')}
+                {inputText('hive_scale_id', 'If you have an automated scale associated with this colony, please enter the hive scale brand and ID number here.')}
                 {this.inputSelect('colony_alive', 'Is the colony alive?', [
                     ['true', 'Yes'],
                     ['false', 'No'],
--- a/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
@@ -27,8 +27,7 @@ class MonthlySurveyColonyForm extends Component {
     inputFactory(inputType) {
         const { onChange, data } = this.props;
         const disabled = data.id !== null;
-
-        return (key, label, required, tooltipDescription) => {
+        return (key, label, required, tooltipDescription, placeholder, pattern) => {
             const tooltip = tooltipDescription
                 ? <Tooltip description={[tooltipDescription]} />
                 : null;
@@ -47,6 +46,8 @@ class MonthlySurveyColonyForm extends Component {
                         onChange={onChange}
                         disabled={disabled}
                         required={required}
+                        placeholder={placeholder}
+                        pattern={pattern}
                     />
                 </div>
             );
@@ -172,11 +173,12 @@ class MonthlySurveyColonyForm extends Component {
                 </div>
             </>
         );
-
+        /* eslint-disable-next-line no-useless-escape */
+        const dateRegex = '\\d{4}-\\d{2}-\\d{2}';
         return (
             <>
                 {inputText('colony_name', 'Colony Name (required)', required)}
-                {inputDate('inspection_date', 'Date of Inspection', required)}
+                {inputDate('inspection_date', 'Date of Inspection', required, false, 'YYYY-MM-DD', dateRegex)}
                 {inputText('hive_scale_id', 'If you have an automated scale associated with this colony, please enter the hive scale brand and ID number here.')}
                 {this.inputSelect('colony_alive', 'Is the colony alive?', [
                     ['true', 'Yes'],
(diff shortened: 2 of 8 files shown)


src/icp/apps/beekeepers/models.py
```diff
@@ -349,7 +349,7 @@ class MonthlySurvey(models.Model):
         null=True,
         blank=True,
         help_text='If you have an automated scale associated with this '
-                  'colony, please enter the hive scale ID number here.')
+                  'colony, please enter the hive scale brand and ID number here.')
```
rajadain: This will need a migration:

```bash
$ ./scripts/manage.sh makemigrations --dry-run --verbosity 3

+ ARGS='makemigrations --dry-run --verbosity 3'
+ [[ 4 -eq 1 ]]
+ vagrant ssh app -c 'cd /opt/app && envdir /etc/icp.d/env ./manage.py makemigrations --dry-run --verbosity 3'
Migrations for 'beekeepers':
  0014_auto_20190611_1116.py:
    - Alter field hive_scale_id on monthlysurvey
Full migrations file '0014_auto_20190611_1116.py':
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('beekeepers', '0013_update_queen_source_options'),
    ]

    operations = [
        migrations.AlterField(
            model_name='monthlysurvey',
            name='hive_scale_id',
            field=models.CharField(help_text='If you have an automated scale associated with this colony, please enter the hive scale brand and ID number here.', max_length=255, null=True, blank=True),
        ),
    ]
```

(with the migration renamed to something meaningful rather than `auto_20190611_1116`)

rajadain:
I have a monthly survey filled with the first colony listed as `A` and the second as `C`, but in prefililng subsequent surveys the order is reversed:

rajadain:
Apart from the reversal, other behavior seems correct.

fungjj92:
Ready for a full look.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
```diff
@@ -128,7 +129,6 @@ class MonthlySurveyColonyForm extends Component {
         } = this.props;
 
         const inputText = this.inputFactory('text');
-        const inputDate = this.inputFactory('date');
```
rajadain: Since Safari and IE don't recognize `input type="date"`, they fallback to `input type="text"`. We don't need this to explicitly be a text field for them. But by making it so, we lose the date picker in the browsers that do have it (such as Chrome and Firefox).

Prefer something like this, which keeps the date picker in modern browsers:

```diff
diff --git a/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx b/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
index 0d831b81..bf41c15c 100644
--- a/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
@@ -129,6 +129,7 @@ class MonthlySurveyColonyForm extends Component {
         } = this.props;
 
         const inputText = this.inputFactory('text');
+        const inputDate = this.inputFactory('date');
         const inputNumber = this.inputFactory('number');
 
         // The first colony is always requried. Henceforth, the colony name
@@ -179,7 +180,7 @@ class MonthlySurveyColonyForm extends Component {
         return (
             <>
                 {inputText('colony_name', 'Colony Name (required)', required)}
-                {inputText('inspection_date', 'Date of Inspection', required, false, 'YYYY-MM-DD', dateRegex)}
+                {inputDate('inspection_date', 'Date of Inspection', required, false, 'YYYY-MM-DD', dateRegex)}
                 {inputText('hive_scale_id', 'If you have an automated scale associated with this colony, please enter the hive scale brand and ID number here.')}
                 {this.inputSelect('colony_alive', 'Is the colony alive?', [
                     ['true', 'Yes'],
```

src/icp/apps/beekeepers/js/src/components/ApiarySurveyListing.jsx
```diff
@@ -22,9 +23,14 @@ const ApiarySurveyListing = ({ apiary, surveys }) => {
     );
 };
 
+ApiarySurveyListing.defaultProps = {
+    lastMonthlySurvey: {},
```
rajadain: Don't need to change this, but it felt a little unusual to use `{}` as the default value rather than a `null`. Is this advantageous in any way?

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
```diff
@@ -168,6 +168,31 @@ class MonthlySurveyForm extends Component {
...
+        // autofill select fields with latest monthly survey
+        if (lastMonthlySurvey.id && !id) {
+            getOrCreateSurveyRequest({
+                apiary: lastMonthlySurvey.apiary,
+                id: lastMonthlySurvey.id,
+            }).then(({ data }) => {
+                const { monthlies: formMonthlies } = this.state;
+                const sortedMonthlies = Object.assign([], data.monthlies)
+                    .sort((a, b) => a.id - b.id); // In ascending order of id
+                const updatedFormMonthlies = formMonthlies.map((monthly, idx) => {
```
rajadain: The code here looks like it should be updating all surveys, even filled out ones. But that's not happening. Consider adding some comments here to make the logic clearer.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
```diff
@@ -37,11 +38,18 @@ const SurveyCard = ({ apiary, dispatch, surveys }) => {
     } else {
         const shownSurveys = surveys.slice(0, 4);
 
+        const sortedMonthlySurveys = surveys
+            .filter(survey => survey.survey_type === SURVEY_TYPE_MONTHLY && !!survey.completed)
+            .sort(survey => survey.month_year);
```
rajadain: This is working correctly, but I'm having a hard time understanding how. This looks like it should sort by ascending order by default, and we're picking the first element in the next line. But that would be the oldest survey, not the most recent one.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyForm.jsx
fungjj92: I can clarify the top-level comment at line 172. This whole `if` block, starting at line 173, only applies if there's a lastMonthlySurvey provided and the survey is blank. With an `id` we get the disabled filled out survey. Without requisite information, we get a fully blank survey.

src/icp/apps/beekeepers/js/src/components/MonthlySurveyColonyForm.jsx
fungjj92: I had thought about this and went with a text input to give a universal experience across browsers. I did a quick consult with design and got: "Generally we try to give a universal experience across browsers. In some cases where time / cost is an issue, we may not...the ideal is to use the “proper” input over something custom, but something custom over a potential downgrade / unexpected experience"

For this reason I'll shift back to a Date input, which still shows up as a text input for the browsers that don't support the date type input.

src/icp/apps/beekeepers/js/src/components/SurveyCard.jsx
fungjj92: I can see how you one would see the sort as counterintuitive. Anyway, it is superfluous, I will remove it.

fungjj92:
Thanks for taking a look! Updated.

rajadain:
Taking another quick look to ensure it still works.

Issue #522, opened 2019-07-11
Update VRTs on s3 with new raster data for several states


fungjj92:
The client will provide us new processed raster data for the states WI, MI, OH, NY, WV. We will want to upload the rasters to their respective staging and production folders in S3 and use GDAL to generate/update the VRTs to include said rasters. 

The command on GDAL is like:
`gdalbuildvrt <destination_vrt_name> <file1> <file2>`

It may be quicker/straightforward enough to copy & replace sections of the existing VRT with references to the new states.

mmcfarland:
After updating these rasters with new existing state data, we'll have to determine what we do with saved project apiaries. As is, the stored values will be (potentially) out of sync with values determined going forward. We have two options we've used in the path:

1. Do nothing, project values remain point-in-time
2. Wipe out saved project values IF this will result in the values being recalculated the next time a user loads the saved apiaries (this is how it works in MMW & Pollination Mapper)

This can be done on a subsequent issue, but wanted to record it here for reflection when implementing this change.

PR #523, opened 2019-07-12
Prepare app for 5 new states' raster data

fungjj92:
## Overview

The client provided raster data for 5 new states, and updated PA, IL, and IN data too. This PR updates the VRT generation script. Also, I unified the code around the indicator label 'insecticide' formerly 'pesticide'. We had an in-between hybrid solution before to avoid renaming & regenerating data.

Connects #522

### Demo

Data completes a full corridor between NY and MI :D
Screen Shot 2019-07-12 at 4 01 31 PM

Data coming in from all included states, screenshot from my local instance
Screen Shot 2019-07-12 at 3 36 55 PM

Staging:
Screen Shot 2019-07-12 at 4 07 21 PM

### Notes

I updated data in the staging data bucket in s3; staging will not return insecticide data until its code is updated to accomodate `insecticide`. Production will be a bit tricky. Either we upload the new VRTs and layers before the updated code is deployed, or after -- in both cases we'll get no data for the insecticide field for a few minutes. I don't think this is ultimately a huge deal, so long as we flush saved apiary indicator data after all the updates are made to ensure fresh, complete data is shown to users.

## Testing Instructions

Pull down this branch and you should get data for all 8 included states when clicking the map.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,17 +309,17 @@ is provided per state and the file naming scheme represents this as:
 
 `[STATE_ABBR]_[INDICATOR]_[FORAGE_RADIUS].tif`
 
-This results in files named like `PA_pesticide_3km.tif`.
+This results in files named like `PA_nesting_3km.tif`.
 
 In order to support multiple states that come as discrete files, a VRT per layer/radius pair
 has been created using the following steps:
 
 ```bash
-gdalbuildvrt pesticide_3km.vrt PA_pesticide_3km.tif IL_pesticide_3km.tif
+gdalbuildvrt nesting_3km.vrt PA_nesting_3km.tif IL_nesting_3km.tif
 ```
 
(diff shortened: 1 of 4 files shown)


mmcfarland:
There also seems to be a CI failure.

fungjj92:
Thanks for the QC! I relayed the WV data question to @dboyer and cleared out "pesticide" labelled files in the staging bucket.

There's always a CI failure unless the workspace is cleared out before running a job. Just reran with a cleared workspace, and the build is green.

I'm going to wait to merge this until WV followup but meanwhile make a followup card about production deployment with notes on clearing out the `scores` column of the `apiary` table and copying over data from the staging to production s3 bucket.

fungjj92:
WV was confirmed by the client to have mostly 0 values. WV is apparently very forested. Go WV?!

Issue #525, opened 2019-07-25
Logged in Users Whitescreen on Staging


rajadain:
On staging, if you log in to an account that already has some apiary scores cached, we get this:

![image](https://user-images.githubusercontent.com/1430060/61908766-70dd7a80-aefe-11e9-90fe-e7548254dc1c.png)

This is possibly because the data structure in the front-end has changed, but the data in the back-end was not migrated into the new structure.

We can either:

1. Write a data migration that reads every existing row and checks if it is in the new format. If not, it adds / removes keys until it is, and saves it.
2. See if an Apiary with `null` scores has them fetched automatically. If so, clear out all the old scores and they will be populated as users log back in to the new system.

PR #526, opened 2019-08-01
Add data migration to nullify scores

fungjj92:
## Overview

The schema of the underlying scores data changed, thus invalidating the cached apiary scores. By wiping the scores, we allow a user to pull fresh data on next login. Said data is not automatically pulled because of budget constraints.

Connects #525 

### Demo
Logged in with wiped data:
Screen Shot 2019-08-01 at 10 27 24 AM

## Testing Instructions

Pull down branch
Try to login and see that the app breaks
Apply the migration `./scripts/manage.sh migrate`
Refresh app and see the screenshot
Pull data and see that it comes back successfully

### DIFF
--- a/src/icp/apps/beekeepers/js/src/actions.js
+++ b/src/icp/apps/beekeepers/js/src/actions.js
@@ -59,7 +59,9 @@ export function fetchApiaryScores(apiaryList, forageRange) {
         }
 
         // check if data for the apiaries already exist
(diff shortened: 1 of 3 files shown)


rajadain:
This is failing for me with the message "can't convert null to object"

image

image

Should we reset it to be `{}` instead?

fungjj92:
Great! Phew ha.

Issue #542, opened 2019-10-24
More Minor Changes to Surveys


rajadain:
- Add “N/A – no colonies lost” option in Spring survey for the “What do you think was the most likely cause of colony loss?” question

PR #543, opened 2019-10-28
Add an 'n/a' option to a spring survey question

fungjj92:
## Overview

The question "What do you think was the most likely cause of colony loss?" suggests that there was colony loss. It is possible there wasn't, and so add an answer option that indicates as such.

This change was tucked away in the client's wires and got missed in #533 

Connects #542 

### Demo

Screen Shot 2019-10-28 at 12 37 34 PM

Screen Shot 2019-10-28 at 12 57 43 PM

## Testing Instructions

Pull down the branch. If your haven't pulled down develop in the last few days, make sure to run migrations and possibly `./scripts/beekeepers.sh install`. Though, there were no added migrations or yarn dependencies here.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { func } from 'prop-types';
 
 import { fetchUserApiaries, flashSuccessModal } from '../actions';
-import { SURVEY_TYPE_APRIL, COLONY_LOSS_REASONS } from '../constants';
+import { SURVEY_TYPE_APRIL, SPRING_COLONY_LOSS_REASONS } from '../constants';
 import {
     arrayToSemicolonDelimitedString,
     getOrCreateSurveyRequest,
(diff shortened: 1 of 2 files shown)


src/icp/apps/beekeepers/js/src/components/AprilSurveyForm.jsx
```diff
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { func } from 'prop-types';
 
 import { fetchUserApiaries, flashSuccessModal } from '../actions';
-import { SURVEY_TYPE_APRIL, COLONY_LOSS_REASONS } from '../constants';
+import { SURVEY_TYPE_APRIL, SPRING_COLONY_LOSS_REASONS } from '../constants';
```
rajadain: Consider changing `COLONY_LOSS_REASONS` rather than making a new `SPRING_COLONY_LOSS_REASONS`. Monthly surveys (the other thing that uses `COLONY_LOSS_REASONS`) are going away anyway, and even if they remain, it is better to keep things in sync. Reduces cognitive load for both developers and users.

fungjj92: We don't want this 'n/a' option for the Monthly Surveys and if we use the same array and end up bringing monthly surveys back, monthly surveys will be incorrect.

rajadain: Hmm, because in the Monthly Surveys this is handled by the "Is this colony alive" question:

image

Alright, let's keep this as it is then.

fungjj92: Indeed! See commit message 4e1c39d

fungjj92:
Thanks for review! 🤞 we've caught everything now

Issue #548, opened 2019-10-30
Map Doesn't Update on New Search if Zoomed In All The Way


rajadain:
If you are zoomed in all the way, searching for a new location does not update the map, until you zoom out one level, then it updates the map:

![2019-10-30 14 35 04](https://user-images.githubusercontent.com/1430060/67888610-d192d980-fb23-11e9-9968-85647a18ef21.gif)

mmcfarland:
We will want to set up a `hotfix/1.4.1` branch so we can do a hotfix release. 

@rajadain or @fungjj92 can you coordinate with operations to make sure this gets scheduled.

PR #549, opened 2019-10-30
Bugfix: unify on map center format

fungjj92:
## Overview

Fix geocoder searches panning & zooming the map. This bug was introduced by formatting coordinate information as an array instead of an object 😱 

Also, I removed the beekeepers prototype because we don't need that anymore and it's generating dependabot PRs.

Refs #548 

### Demo

bees

## Testing Instructions

Use the geocoder and see that selecting a location pans & zooms the map as many times as you select new locations.
See no weird behavior clicking apiary card markers to pan & zoom to their locations.

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/ApiaryCard.jsx
@@ -166,7 +166,7 @@ class ApiaryCard extends Component {
 
     zoomPanMapToLocation(lat, lng) {
         const { dispatch } = this.props;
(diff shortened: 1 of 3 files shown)


rajadain:
Taking a look now

rajadain:
+1 tested, this is working correctly. However, this is too big a change for a hotfix. We should separate this into two PRs:

1. A branch off `develop` that targets `develop` which removes the prototype
2. A branch off `master` that targets `hotfix/1.4.1` which fixes the map zoom

fungjj92:
Thanks for taking a look. Can do.

Issue #560, opened 2020-02-05
Change button for editing apiary names


dboyer:
Replace the check mark that is visible near the apiary name when in edit mode with a button specifically including the word “Save.”

PR #562, opened 2020-02-12
Change button for editing apiary names

TheresaWilkin:
## Overview

Users had trouble interpreting the image buttons for edit and save
within the card component, so we replaced the icons with text to clarify
their purpose. Refactored the CardButton component export into two
higher-order components, CardButtonWithText and CardButtonWithIcon, to
allow reusability of the CardButton code as well as differing sets of
required props for the different button types. The components are
co-located in the CardButton file as named exports.

Connects #560

### Demo

Screen Shot 2020-02-12 at 3 07 39 PM

image

## Testing Instructions

- Check out this branch
- Run ./scripts/beekeepers.sh start
- Add some Apiaries
- Edit their names
- Ensure you see 'Edit' and 'Save' instead of pencil and checkmark icons
- Ensure that editing still works correctly

### DIFF
--- a/src/icp/apps/beekeepers/js/src/components/CardButton.jsx
+++ b/src/icp/apps/beekeepers/js/src/components/CardButton.jsx
@@ -6,6 +6,7 @@ const CardButton = ({
     filled,
     tooltip,
     onClick,
+    text,
 }) => {
     const fillOrOutline = (() => {
         if (filled === true) return '-fill';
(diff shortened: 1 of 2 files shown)


src/icp/apps/beekeepers/js/src/components/CardButton.jsx
```diff
@@ -20,21 +21,47 @@ const CardButton = ({
...
     filled: bool,
     tooltip: string.isRequired,
     onClick: func,
+    text: string,
 };
 
 CardButton.defaultProps = {
     filled: null,
     onClick: () => {},
+    icon: '',
```
rajadain: Consider using `null` instead of empty string here

TheresaWilkin: fixed in d1b8801

Issue #589, opened 2020-05-18
Adjust rankings based on new bins from Beescape team


dboyer:
Waiting on new low, medium, and high rankings.

dboyer:
Low: 79
Medium: 159
High: 226

mmcfarland:
Five point scale from client:

Using the five point scale, the new values would be:
min: 0
low: 79
mid: 159
high: 226
max: 1271

PR #591, opened 2020-05-18
Update insecticide reference ranges

mmcfarland:
## Overview

The value ranges for insecticide were recalibrated against a national
data layer. The fence-post values (0, 1000) were chosen to present
arbitrarily low and high values. The middle values of 79 and 226
represent the thresholds for low and high values, with medium
represented between the two.

Connects #589 

### Demo

image

### Notes

I had to interpret the values provided in the connected issue into the format required by the application, but this was confirmed with follow up to the client

## Testing Instructions

* Test 3 and 5km values across the CONUS, and verify that the insecticide values are labeled appropriately based on this new scale. High values can be found just south of Yuba City, CA


mmcfarland:
Update: I updated the high value from 1000 -> 1271 per instruction from the client.

